PR #2, opened 2023-02-20
agent Manager

peppicus:
prima bozza


peppicus:
i added a class as agent manager

peppicus:
fix vari

PR #25, opened 2023-03-15
Automate Python workflow using pre-commits

umbertogriffo:
## How it works

Automate code reviewing formatting

## How to use it

The code style guide can be enforced through `pre-commit` hooks. When cloning/setting up a new repo you will need to set up the hooks.

1. Download `pre-commit` by running `pip install pre-commit`
2. In the root of the repository run:
   ```commandline
   pre-commit install --hook-type pre-commit
   ```
Pre-commit runs when you commit files, if the formatting is not black-compliant or the code fails linting, the commit will be blocked.
In the case of formatting `black` will have run, and you will need to re-add the file(s), in the case of `flake8` you will need to sort out any linting issues and re-add the files to commit.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,36 @@
+default_stages: [ commit ]
+repos:
+   # pre-commit is also configured to format yaml, trim whitespace, add new lines,
+   # check for merge conflicts, and checks for python debug statements.
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+      - id: check-yaml
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+      - id: check-merge-conflict
+      - id: debug-statements
+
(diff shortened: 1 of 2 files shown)


pieroit:
Hi @umbertogriffo thank you for this.
Can you confirm the local setup is optional?
What happens if someone working on a fork commits but has no `pre-commit` installed?

I want people to submit PRs even without having installed `pre-commit`, but I will set it up and use it as you describe.

umbertogriffo:
Hey @pieroit really glad to contribute!

Yes, It's totally optional. Nothing will happen if someone working on a fork commits has no `pre-commit` installed (Alghtout would be easy to have by default if we add a `pipfile` to manage the Python environment).

In fact, the only way you have to force people to adopt the code style is also to have a `validation` step in the GitHub Action triggered when you open a PR.

Maybe we can also think of implementing CI/CD in the future. I could also help with that.

umbertogriffo:
No worries mate! :D I'll be here when will be the right time.

PR #31, opened 2023-03-18
feat(24): Introduces voice input

### LINKED ISSUE #24: Voice Input and Text Transcription for Cheshire Cat Interaction
As a user, I would like to interact with the Cheshire Cat using voice. 
This requires adding voice input capability to our web application. 
The recorded audio will then be transcribed into text using the Web Speech API browser API. 
The resulting text will be sent to the backend for processing.

To provide a smooth user experience, we aim to implement a WhatsApp-like UX interaction for the Cheshire Cat. 
This will allow users to easily interact with the cat using their voice directly and see the corresponding text responses in real-time

antonioru:
Introduces voice messaging with the cat! 

closes #24


pieroit:
Thanks this is awesome

PR #36, opened 2023-03-21
Feat: Setting Table stored on SQlite with SQLAlchemy

umbertogriffo:
Here are the interfaces that we can use to do all the CRUD related to the settings.
The table schema is just for example purposes. We can change it as needed.

Changes made:
- Added a new file `database.py` which creates the database engine and session for SQLAlchemy.
- Added a new file `models.py` which defines the SQLAlchemy models for the settings table.
- Added a new file `schemas.py` which defines the Pydantic schemas for the Setting model.
- Modified `main.py` to create the settings table in the SQLite database.
- Added `setting.py` which defines the endpoints to the CRUD.

### How to test:

Use `curl` or `Postman` to send HTTP requests to the API endpoints, or use the `http://0.0.0.0:8082/docs` page directly.
Curl example:
```
curl -X 'POST' \
  'http://0.0.0.0:8082/settings/' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "name": "api-key",
  "value": "xxxx",
  "category": "secure",
  "enabled": true,
  "createdAt": "2023-03-21T15:26:35.909Z",
  "updatedAt": "2023-03-21T15:26:35.909Z"
}'
```


pieroit:
Hi @umbertogriffo thanks.

I'm reviewing and have a few questions:
- since you are defining both SQLAlchemy models and Pydantic schemas, why not using directly SQLModel? https://sqlmodel.tiangolo.com/
- it is not clear to me how to connect LLM settings (and incoming Embedder and VectorStore settings) with this settings system? See here: https://github.com/pieroit/cheshire-cat/blob/main/web/cat/config/llm.py
- How do you suggest we orchestrate between settings via .env, settings via db and settigns via plugin? I am thinking about this priority order (from high priority to low): plugin > db > env

Thanks

pieroit:
or maybe use the db as a source of truth and let the plugins read/write the db

umbertogriffo:
Anytime @pieroit!

> since you are defining both SQLAlchemy models and Pydantic schemas, why not using directly SQLModel? https://sqlmodel.tiangolo.com/

It's a good idea, I wasn't aware of that. Do you wanna make me that in this PR? 

> it is not clear to me how to connect LLM settings (and incoming Embedder and VectorStore settings) with this settings system? See here: https://github.com/pieroit/cheshire-cat/blob/main/web/cat/config/llm.py

For instance, Let me think...
To configure `llm.py`, what we would need is the following info:
 
| id  | name  | value  | category  | sub_category  | enabled |
|---|---|---|---|---|---|
| aa | openai-api-key  | xxxx  | llm  | openai  | True |
| bb | cohere_api_key  | yyyy  | llm  |  cohere | False |
| cc | hugging_model_id  | zzzz | llm | hugging | False |
| dd | hugging_repo_id |  wwww | llm | hugging | False |
| ee | hugging_api_token | tttt  | llm | hugging | False |
| ff | hugging_endpoint_url | URL  | llm | hugging | False |

In [CheshireCatSettings](https://github.com/pieroit/cheshire-cat/blob/main/web/cat/config/settings.py), we could use the queries to get the setting for the specifically enabled object filtering by `category`, `sub_category` and `enabled` equal to true. So we could do the same for the Embedder and Vectore Store.

> or maybe use the db as a source of truth and let the plugins read/write the db

I think it would be cleaner.

umbertogriffo:
> I was trying to do it myself, but I just git stashed so you can do it better :D

No worries! :D I've moved to SQLModel successfully.

> But then it messes up with pydantic BaseSetting/BaseModel.
Maybe we should take more time and think about this. Are you available for a call?

Regarding the dot notation suggestion, could you please elaborate on `it messes up with pydantic BaseSetting/BaseModel`?.

Btw maybe we could also consider using a hierarchical structure. This could be achieved by adding additional columns to the table, such as:

* **parent_id**: a reference to the id of the parent configuration, indicating that this configuration is a sub-configuration of the parent configuration. If a configuration has no parent, this column would be null.
* **level**: an integer indicating the depth of the configuration in the hierarchy. The top-level configurations would have a level of 0, and their direct children would have a level of 1, and so on.

With these additional columns, we could represent a hierarchical structure of configurations. 

For example:

id | name | value | category | enabled | parent_id | level
-- | -- | -- | -- | -- | -- | --
1 | openai-api-key | xxxx | llm | True | null | 0
2 | openai-temperature | yyyy | llm | True | 1 | 1
3 | cohere-api-key | zzzz | llm | False | null | 0
4 | hugging-model.repo | tttt | embedder | True | null | 0
5 | plugin-name| uno | plugins | True | null | 0
6 | plugin-custom-setting-1 | 10 | plugins | False | 5 | 1
7 | plugin-custom-setting-2 | 20 | plugins | False | 6 | 2
8 | plugin-name| due | plugins | True | null | 0
9 | plugin-custom-setting-1 | 100 | plugins | False | 5 | 1
10 | plugin-custom-setting-2 | 200 | plugins | False | 6 | 2

In this example, the configurations with id 1, 3, 4, 5, and 8 are top-level configurations, while the configurations with id 2, 6, 7, 9 and 10 are sub-configurations of their parent configurations. The level column indicates the depth of each configuration in the hierarchy, with the top-level configurations having a level of 0 and their direct children having a level of 1, and so on.

This structure would allow us to represent configurations of arbitrary depth. We could also use the parent_id and level columns to traverse the hierarchy of configurations if needed.

> Are you available for a call?

Tomorrow I should have time. I'll ping you on Linkedin.

> The cat will expose hooks and utility functions for plugins to read and write in the db. If there are no plugins, we can use defaults described in CheshireCatSettings or (better) delegate them to the quickstart_cat_plugin (which is there to experiment on the plugin system and can be shipped with the cat by default).

The second option sounds good!

umbertogriffo:
Hey @pieroit should we merge it and use it later on as we agreed?

pieroit:
@umbertogriffo I'm doing the merge right now! Please do not add to the PR

Will push after trying it out a little... thanks!

PR #42, opened 2023-03-23
Advanced way

Mte90:
Right now it is just without docker for python I have to test it also for the frontend.
I just bumped a dependence as I have python 3.11 and added support for `.env` file that automatically python will read as environment.


Mte90:
Everything works but there are troubles with css.

immagine

My guess is that something in vue is not working as the css as it is compiled in this way:
```
.ccat-Toolbar?used-toolbar .ccat-Toolbar?used-title h1 {
 font-family:var(--app-title-font-family);
 font-weight:var(--app-title-font-weight);
 letter-spacing:var(--app-title-letter-spacing);
 line-height:var(--app-title-line-height);
 font-size:inherit;
 margin:0;
 font-size:var(--app-text-size-1)
}
```
Creates troubles, ? isn't supported on selectors.

But my knowledge in vue is 0.

Mte90:
Changed the npm command and now the css are loaded rightly.

Mte90:
Aligned with the latest changes.

The recap:

* support for env file directly from python instead as environment variable loaded by docker. In this way can support the "advanced way" without docker but also more environment variables
* in readme moved below the "advanced way" to use it without docker

Issue #49, opened 2023-03-25
Add hook for summarization


calebgcc:
It's useful to have inside CheshireCat a way to summarize text/documents before saving embeddings. 
We can add a new hook in the default plugin and make it available in CheshireCat class. Something like this maybe:

```python
def load_plugins(self):
  ...
  self.embedder = ...
  self.summarizer = self.mad_hatter.execute_hook("get_language_summarizer", self)
  ...
```
We can use Hugging Face maybe for the default one? let me know :)

pieroit:
> It's useful to have inside CheshireCat a way to summarize text/documents before saving embeddings.

That's a great idea, summarization could be useful for:
- saving in memory conversation chunks and not only user utterances / documents
- when a document is ingested, we should also memorize summaries and not only simple text chunks
- when the prompt or any of its contents gets too large, summarize parts of it

So I'll propose you this:
- let's introduce a method for summarization in the [main Cat class](https://github.com/pieroit/cheshire-cat/blob/main/web/cat/looking_glass.py), similar to present methods like `get_hyde_embedding` or `recall_memories_from_text`.
- we than close the PR and reason on how to design all of this methods to be pluggable as hooks, with a good default in the plugin folder

Sounds good? Any more idea?

> We can use Hugging Face maybe for the default one? let me know :)

We just need a summarizer chain on the already present LLM, with a default prompt.
A good example is here with [HyDE embedding](https://github.com/pieroit/cheshire-cat/blob/main/web/cat/looking_glass.py#L139).
Take a look at [langchain chains](https://langchain.readthedocs.io/en/latest/use_cases/summarization.html) to see what is possible.
I'm here for any question! Thanks a lot

calebgcc:
Sound good!

PR #52, opened 2023-03-25
Feature/add summarization

calebgcc:
First iteration to close issue #49.

After studying a bit the [documentation](https://langchain.readthedocs.io/en/latest/modules/indexes/chain_examples/summarize.html) I decided to use "map_reduce" as a default chain_type that should scale well with large files.

I tested the method with a small .txt file (~800 words) and it worked smoothly.
I tried also to feed [alice.txt](https://github.com/pieroit/cheshire-cat/blob/7a242310f2e1dbb40455502c68dd8b21b82679de/alice.txt) but I got the following error:
```
openai.error.InvalidRequestError: This model's maximum context length is 4097 tokens. 
However, your messages resulted in 37730 tokens. Please reduce the length of the messages.
```

We should discuss a method to adopt in this or in a future PR to summarize large files without exceeding the maximum context length of llm. Also, right now the method accepts only documents but maybe we may need a more general summarizer that gets text in input (maybe text from the chat). We can opt to use a simple LLMChain for that purpose.

### DIFF
--- a/web/cat/looking_glass.py
+++ b/web/cat/looking_glass.py
@@ -1,6 +1,7 @@
 import time
 
 import langchain
+from langchain.chains.summarize import load_summarize_chain
 from cat.utils import log
 from cat.memory import VectorStore, VectorMemoryConfig
 from cat.db.database import get_db_session, create_db_and_tables
@@ -65,6 +66,9 @@ class CheshireCat:
         self.hypothetis_chain = langchain.chains.LLMChain(
             prompt=hypothesis_prompt, llm=self.llm, verbose=True
         )
+
+        # TODO: import chain_type from settings
+        self.summarization_chain = load_summarize_chain(self.llm, chain_type="map_reduce")
 
         # TODO: can input vars just be deducted from the prompt? What about plugins?
         self.input_variables = [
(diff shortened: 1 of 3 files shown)


pieroit:
Good job!

Maybe the reason  the prompt gets so long is because [load_summarize_chain](https://github.com/hwchase17/langchain/blob/master/langchain/chains/summarize/__init__.py#L111) runs a MapReduceDocumentsChain, which summarizes each document independently and then stuff the summaries into the output prompt. ([here the responsible code](https://github.com/hwchase17/langchain/blob/master/langchain/chains/summarize/__init__.py#L40))

So it basically expects a list of long documents as inputs, and not a long list of short documents as we do. XD

> right now the method accepts only documents but maybe we may need a more general summarizer that gets text in input (maybe text from the chat). We can opt to use a simple LLMChain for that purpose.

Agree on a simple custom summary chain, reused hierarchically llama-index style (see below).
Starting from a list of short sentences*:
- summarize independently docs that are too long (just in case the splitter did not work)
- group the documents n by n (respecting sequence) and make group summaries
- run the same on "supergroups"
- repeat until the top summary is SHORT.
All of the summaries in the tree get vectorized and pushed into memory along the original sentences.

An alternative is to use a specific library for this like [llama-index](https://gpt-index.readthedocs.io/en/latest/reference/indices/tree.html) which already has tree summaries. It is also [compatible with langchain](https://gpt-index.readthedocs.io/en/latest/how_to/custom_llms.html#example-changing-the-underlying-llm)

Tell me what you think, I may be wrong. Thank you for the PR

*the text splitting done in rabbit_hole is trivial and should be substituted with a langchain TextSplitter.

**P.S.: let's write the custom prompt for the custom `summary_chain` in the code so it can be moved to plugins (cool to have an hook to just define the summarization prompt, and another hook for totally custom summarizer_chain).

PR #55, opened 2023-03-26
.env FILE

alessandro-Petrone:
for those who don't have the gpt API, they can use mine


pieroit:
You are very generous @alessandro-Petrone but I don't think it is a good idea :
- someone may abuse the key to do something else
- if you delete the key, than the repo contains an invalid key
- the .env file is private, should not stay on the repo

Thank you again, I understand and appreciate, but can't be done :*

PR #67, opened 2023-03-30
misc(backend): sort imports, cors origins from .env and readme update

pymike00:
Hi everyone,

Just a quick PR to offer a few improvements, mostly to the Python backend

1. Python imports sorted [according to PEP8 ](https://peps.python.org/pep-0008/#imports).
2. Values for the cors origins list in main.py are now [taken from .env](https://github.com/theskumar/python-dotenv), as requested in your TODOs.
3. README.md has been updated with a "Docs and Resources" section, where a link to Piero's YouTube presentation has been added.

Best,
M.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,12 @@ docker-compose down
 
 <p align="right">(<a href="#readme-top">back to top</a>)</p>
 
+
+## Docs and Learning Resources
+
+- [YouTube Presentation: Meet the CHESHIRE CAT - Open Source AI](https://youtu.be/srsaYy0xmkc)
+
+
 ## Roadmap
 
 - QA / tests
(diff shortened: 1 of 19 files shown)


pieroit:
Thank you @pymike00 !! Appreciated.

I removed the `python-dotenv` dependency because I prefer to only have a root level `.env` for docker and enforce docker usage instead of a local setup. I am a little rigid about this but these are the reasons:
- avoid issues and PRs about local setups (for some people venv is broken, for some other npm, and who knows what else on three different operating systems :D)
- being free to add/remove microservices without worrying about ad-hoc scripts (there was a Qdrant db, then we removed it, and now we are realizing it is necessary again)
- don't want to split docs and tutorials in with/without docker versions, at least not until the project is stable enough.

Thanks again, a big PURRR from the cat

Issue #16, opened 2023-03-09
prompts get too large


pieroit:
Use more deeply langchain routines to keep the prompt at limited length (CombineDcoumentsChain etc.).
Summarization may also be appropriated when documents are uploaded.

calebgcc:
I closed issue #49 since this issue is more specific, we can use this issue to discuss how to further implement summarization 🙌.

About your comment in PR #52:

 I can test other [chain_type](https://python.langchain.com/en/latest/modules/indexes/chain_examples/summarize.html) to see if I get the same problem with large files.

I'm going to dig a little deeper into the docs that you left (about llama-index) to understand better how to implement the custom summary chain, but if I understand correctly the basic idea is:
- get a list of strings in input
- group them in different docs
- get summary from docs (which becomes new input)
- repeat until we have one single short summary

pieroit:
@calebgcc I introduced in the rabbit_hole a TextSplitter wich can be customized (chunk_size and chunk_overlap). So Cat users can decide themselves how long they want their text chunks.

You find in `docs` [here](https://github.com/pieroit/cheshire-cat/blob/main/web/cat/rabbit_hole.py#L53) a list of [langchain Documents](https://python.langchain.com/en/latest/modules/indexes/document_loaders/examples/copypaste.html) (which is just an object with text and metadata) to experiment with file summarization.

pieroit:
PR #68 merged and now file uploads do summarization.
Next step is do summaries when the list of memories recalled [here](https://github.com/pieroit/cheshire-cat/blob/main/web/cat/looking_glass/cheshire_cat.py#L199) makes the prompt too large.
Leaving this issue open

calebgcc:
@pieroit I was trying to trigger this error, but I think summarizing and chunking the documents solved it.

The documents that are retrieved from the cat are often too small to cause problems, and this adds up to the fact that k is by default 5.

Maybe prompt summarization is no longer necessary, let me know how to proceed, for example we can try increasing the value of k to see how it affects the prompt.

pieroit:
Increasing k is a good test, also if somebody uploads a doc and chooses a large chunk size the problem remains.

There should be a check before inserting memories in the prompt, if they are "too long" they should be summarized.

We can postpone the problem and close this issue as we are mostly covered, or if you feel like it also tackle the above.

Thanks 🙏

PR #68, opened 2023-03-31
Feature/summarization

calebgcc:
Another step forward to solve #16.

I added a prompt for the default summarization method that can be imported from plugins (also changed chain type from "map_reduce" to "stuff").

I tested the new text_splitter and it works great, the problem of reaching maximum token is solved now.
We can talk about passing a group of documents instead of just one document, this could help implement hierarchical summarization very easily (if we iterate this step on the output).

### DIFF
--- a/web/cat/looking_glass/cheshire_cat.py
+++ b/web/cat/looking_glass/cheshire_cat.py
@@ -69,7 +69,7 @@ class CheshireCat:
         )
 
         # TODO: import chain_type from settings
-        self.summarization_chain = load_summarize_chain(self.llm, chain_type="map_reduce")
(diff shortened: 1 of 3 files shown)


pieroit:
Yes @calebgcc go for it!

calebgcc:
@pieroit I just added iterative summarization.
It works fine, the only issue now is that for a long document it takes ages, I tried with [alice.txt](https://github.com/pieroit/cheshire-cat/blob/25671605eb9f85aea856c8bd7f3f1d0713dfa272/alice.txt)  and it took roughly 20 minutes.

Two possible improvements could be:
- saving intermediate results
- use threads (or similar) to speed up the process

PR #71, opened 2023-04-03
Improve logs

peppicus:


pieroit:
Hi @peppicus sorry for the delay.
Can you please:
1 - separate the logs PR from the Hooks PR
2 - try a different way to add arguments to the hooks, I don't think the global variable does the job (I'll leave you comments in the code)

web/cat/mad_hatter/decorators.py
```diff
@@ -7,11 +7,20 @@
 
 
 log(tool)
+hooks = []
```
pieroit, line 10: This does not look good to me. Hooks array should live inside the MadHatter and Hooks do not need to know that they are part of an array

web/cat/mad_hatter/decorators.py
```diff
@@ -7,11 +7,20 @@
...
+        def hook_func(*args, **kwargs):
+            # log(func)
+            return func(*args, **kwargs)
 
-def hook(func):
-    def cat_hook_wrapper(*args, **kwargs):
-        # log(func)
-        return func(*args, **kwargs)
-
+        hooks.append({
```
pieroit, line 18: We can write the hook decorator as a class and having this information as instance attributes.
So in the plugins hooks are simply functions, but we are more free on the hook decorator level.
What do you think?

PR #87, opened 2023-04-08
Added custom summarization chain

calebgcc:
I added a very simple custom summary chain.
Right now is possible to decide which one to use through a parameter in the method get_summary_text(...) but after some tests, I saw that they are equal in time and results, we can opt to leave only the custom chain.

The only pro of using the langchain summarization function could be tuning chain_type.


pieroit:
Great! Thanks

PR #110, opened 2023-04-12
Refactor: RabbitHole is made a component of CheshireCat to allow auto-ingest

nicola-corbellini:
The RabbitHole is now a class embedding the function "ingest_file".
CheshireCat class creates one RabbitHole instance at initialization time and can call "ingest_file".
Upload.py has been updated to make the ingestion works both from GUI and scripts

### DIFF
--- a/web/cat/looking_glass/cheshire_cat.py
+++ b/web/cat/looking_glass/cheshire_cat.py
@@ -22,7 +22,7 @@ class CheshireCat:
         self.bootstrap()
 
         # Rabbit Hole Instance
-        self.rb_hole = RabbitHole()
+        self.rabbit_hole = RabbitHole()
 
     def bootstrap(self):
         """This method is called when the cat is instantiated and
@@ -234,10 +234,10 @@ class CheshireCat:
         """
 
         # Load the file in the cat memory
-        self.rb_hole.ingest_file(self,
-                                 file=file,
-                                 chunk_size=chunk_size,
-                                 chunk_overlap=chunk_overlap)
+        self.rabbit_hole.ingest_file(self,
+                                     file=file,
+                                     chunk_size=chunk_size,
+                                     chunk_overlap=chunk_overlap)
 
     def __call__(self, user_message):
         if self.verbose:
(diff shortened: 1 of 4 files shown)


pieroit:
Great jobs thanks!
There are a few things to fix before merge, please see comments in the PR

web/cat/looking_glass/cheshire_cat.py
```diff
@@ -18,6 +21,9 @@ def __init__(self, verbose=True):
         # bootstrap the cat!
         self.bootstrap()
 
+        # Rabbit Hole Instance
+        self.rb_hole = RabbitHole()
```
pieroit, line 25: please call the attribute `self.rabbit_hole` as it is more readable

web/cat/routes/upload.py
```diff
@@ -15,11 +14,6 @@ async def rabbithole_upload_endpoint(
     request: Request,
     file: UploadFile,
     background_tasks: BackgroundTasks,
-    chunk_size: int = Body(
```
pieroit, line 18: Add back `chunk_size` and `chunk_overlap` params to the endpoint, so scripts uploading files to the Cat can decide how the text is splitted and overlapped

web/cat/routes/upload.py
```diff
@@ -42,9 +36,7 @@ async def rabbithole_upload_endpoint(
         )
 
     # upload file to long term memory, in the background
-    background_tasks.add_task(
-        ingest_file, ccat, file, chunk_size=chunk_size, chunk_overlap=chunk_overlap
-    )
+    background_tasks.add_task(ccat.send_file_in_rabbit_hole, file)
```
pieroit, line 39: the function call should take `chunk_size` and `chunk_overlap` args as it was before

pieroit: @nicola-corbellini to try out the endpoints see `localhost:1865/docs/`

pieroit (changes requested):
Can you fix this stuff?

web/cat/routes/upload.py
nicola-corbellini: Ouch sorry, I naively misunderstood the code. 
I'll find a way to fix this, as before removing the `chunk_size` and `chunk_overlap` params from the endpoint it raised an error. I'll dig into this!

pieroit:
@nicola-corbellini congrats on your first successful PR!
Many thanks :)

PR #111, opened 2023-04-12
Add optional APIs auth, close #83

### LINKED ISSUE #83: Simple Auth with .env
Let's postpone a full fledged user management as proposed in #62 and go for a simple token auth:
- if there is no `AUTH_TOKEN` in the `.env` all endpoints are public
- it there is one, endpoints only reply to requests having `Authorization: Bearer <auth_token>` in the header

EugenioPetulla:
API_KEY accept multiple keys pipe separated and is optional.

Notice: Swagger changes its behavior depending on whether or not the key is in the .env file. To achieve this we have to keep track of new routes and update the openapi.py file as well or some routes will be flagged as "protected" in the docs also if they are not. 

```.env``` EG.
```bash
API_KEY=key1|key2 | key3 |key4
```

Next step, after pr is merged, write some docs about optional api auth.

### DIFF
--- a/web/cat/api_auth.py
+++ b/web/cat/api_auth.py
@@ -1,0 +1,14 @@
+from fastapi import Header, Security, Depends, HTTPException
+from fastapi.security.api_key import APIKeyHeader
+import os
+
+API_KEY = [key.strip() for key in os.getenv("API_KEY", "").split("|") if key.strip()]
+api_key_header = APIKeyHeader(name="access_token", auto_error=False)
+
+def check_api_key(api_key: str = Security(api_key_header)):
+    if not API_KEY:
+        return None
+    if api_key in API_KEY:
+        return api_key
+    else:
+        raise HTTPException(status_code=403, detail="Invalid API Key")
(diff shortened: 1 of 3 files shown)


pieroit:
Great thanks!

pieroit:
Hi @EugenioPetulla

I was trying to automate the assignation of `security` to `None` on the routes, as you do [here](https://github.com/pieroit/cheshire-cat/blob/main/core/cat/routes/openapi.py#L27).

I was writing something like:

```python
paths = openapi_schema["paths"]
for _, path in paths.items():
    for __, verb in path.items():
        verb["security"] = None # default was [{'APIKeyHeader': []}]
```

But I realized everything works as expected if I do no assignation (all routes are open without `API_KEY`, all are closed with `API_KEY`).

> Notice: Swagger changes its behavior depending on whether or not the key is in the .env file. To achieve this we have to keep track of new routes and update the openapi.py file as well or some routes will be flagged as "protected" in the docs also if they are not.

Can you explain what problem you encountered that forced you to assign None on the security field?

PR #113, opened 2023-04-13
Updated Getting started

nicola-corbellini:
I added some detail in the Getting Started section and commented out the Contribution section as it may be more appropriate in the README. However, if it is intended to stay there I can restore it


pieroit:
Thanks!

PR #114, opened 2023-04-13
Solved markdown upload bug

calebgcc:
I simply added ".md" in the input file tag so that markdown files are visible in the "custom files" picker (no need to use "all files" to upload markdown files now)

Then I rely on guess_type to know the content-type of the file instead of trusting what the browser picked.


calebgcc:
@pieroit no worries, working on that

Issue #103, opened 2023-04-11
Add a WorkingMemory class


pieroit:
Episodic and Declarative memory contents that are relevant and inserted into the prompt should be managed by a separated class `WorkingMemory` so they can be not only used in the prompt, but also in plugins and in the rabbit hole.
This opens up *a lot* of use cases

@calebgcc @nickprock @peppicus  what do you think?

pieroit:
A Working memory as proposed by @nicola-corbellini is not only a good idea to pass short-term lived information around (in both core and plugins), but also coherent with the idea of a cognitive architecture.

We could create a little class `WorkingMemory` in `core/cat/memory/working_memory.py`. An instance of this class is attached as attribute to the cat so we can do `cat.working_memory.something()`

It should encapsulate a dictionary in which both the core and plugins can store arbitrary information. At execution time we can add content to the working memory with something like:

`cat.working_memory.keep_in_mind("custom-key", custom_content)`

There may be utility methods like:

`working_memory.reset()`
`working_memory.forget("custom-key")`

And in any case one should be able to directly manipulate the dictionary with something like

`working_memory["custom-key"] = [1, 2, 3]`

For this latest case, we can rely to python [magic methods](https://www.geeksforgeeks.org/__getitem__-and-__setitem__-in-python/) to make the whole instance act as a dictionary :dancers: 

When this is done I will try to let langchain write the agent steps into the working memory so the prompting cycle is available and the `why` can be accessible from plugins

As a final note, recent [conversation history](https://github.com/pieroit/cheshire-cat/blob/main/core/cat/looking_glass/cheshire_cat.py#L101) should be easily incapsulated in the working memory instead of begin directly attached to the cat. See for example issue #172  

@nicola-corbellini if you comment this issue I'll assign to you, thanks!
@calebgcc also your view on this would be helpful

nicola-corbellini:
I have to correct my last comment: I was reasoning about the implementation and a couple of doubts popped to my mind.

Both these methods:
`working_memory.reset()`
`working_memory.forget("custom-key")`
are already `dict` built-in  methods. Namely, `dict.clear()` and `dict.pop("custom-key")`.

Moreover, you, @pieroit, stated that devs should be able to directly manipulate the `cat.working_memory` instance, that should act like a dictionary. 
Thus, I asked myself: is a `WorkingMemory` class really necessary? 

The answer I gave to myself is to create something as simple as:
```python
class WorkingMemory(dict):
[...]

PR #115, opened 2023-04-13
Small refactor of RabbitHole class

calebgcc:
I split the original method into two methods:
- `file_to_docs` which, given a file chunks it into Documents
- `store_documents` which, given a list of Documents, stores them in memory

The next step is to create a list of documents from an URL, the method used to store them in memory will be the same. (maybe we can refactor further and put the store method inside the cat, let me know)

### DIFF
--- a/core/cat/looking_glass/cheshire_cat.py
+++ b/core/cat/looking_glass/cheshire_cat.py
@@ -234,10 +234,15 @@ class CheshireCat:
         :param chunk_overlap: number of overlapping characters between consecutive chunks
         """
 
-        # Load the file in the cat memory
-        self.rabbit_hole.ingest_file(
-            self, file=file, chunk_size=chunk_size, chunk_overlap=chunk_overlap
+        # split file into a list of docs
+        docs = RabbitHole.file_to_docs(
+            file=file, chunk_size=chunk_size, chunk_overlap=chunk_overlap
         )
+
+        # get summary and store in memory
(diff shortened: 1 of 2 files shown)


pieroit:
Great idea! Thanks

> The next step is to create a list of documents from an URL, the method used to store them in memory will be the same. (maybe we can refactor further and put the store method inside the cat, let me know)

The method to store docs should be part of the `VectorStore`/`Memory` class.
We could discuss a little how to orchestrate `LongTermMemory`, `WorkingMemory` (#103) and `RabbitHole`.
Let's involve also @nicola-corbellini 

This is cooler and cooler

nicola-corbellini:
You guys see far beyond my current horizons, but I'll try follow you!

Honestly, I have not yet clearly understood what the `VectorStore` class does.
Also, isn't the `RabbitHole` a metaphor for the `Memory`? And how do see `LongTermMemory` different than a more generale `Memory`?

What I'm imaginining right now is having a parent class `RabbitHole` (read `Memory`) and two children classes (i.e. `LongTermMemory` and `WorkingMemory`). These memories could embed  `store_documents` and all the current `CheshireCat` ' s methods related to memory usage (i.e. `recall_memories_from_text', 'recall_memories_from_embedding' and 'format_memories_for_prompt').
The `WorkingMemory` could indeed recall memories from episodic and declarative memories with a tunable K to have more strict of fancy results. 

Let me know if I've gone too far or I'm saying trivial things.

PR #127, opened 2023-04-17
Updated Home page and Components description

nicola-corbellini:
Short introduction, updated "Getting started" subsection with correct commands plus a 30s screen recording and a high-level description of the components of the Cat.
If anything is missing or should be further detailed, let me know please. 
In the future I'll describe the main loop.


pieroit:
Thanks!

Issue #125, opened 2023-04-17
Feature Request: Ingest Website content using URL


calebgcc:
It might be helpful if the cat could ingest web content such as blogs or Wikipedia articles through a link.
This should be treated exactly like the ingestion of documents. (parsing, summarization, storing, ...)

pieroit:
@calebgcc agree!
Let's start from a simple URL

pieroit:
It also opens up possibility to make a browser extension for the cat

PR #129, opened 2023-04-18
Added url ingestion endpoint + cat

calebgcc:
Backend implementation to solve #125
Waiting for frontend implementation to close the issue

I added a new endpoint `POST /rabbithole/web/` that accepts an _url_ to ingest.
The process is identical to file ingestion: Loader --> Splitter --> Summary --> Store

I tested it with the Wikipedia article on LLMs using chatgpt.
Many documents inserted in memory are actually useless, they contain Wikipedia information like the table of contents and etc... it shouldn't create too many problems, a way to solve this issue is to store only summaries. 

Screenshot 2023-04-18 alle 12 31 52
Screenshot 2023-04-18 alle 12 31 13


pieroit:
Cool!

PR #137, opened 2023-04-18
WiP: Configuration for Cognitive Services Azure OpenAI

zioproto:


core/cat/plugins/quickstart_cat_plugin/models.py
```diff
@@ -72,20 +72,33 @@ def get_language_model(cat):
...
+                   # model_name: '....'  # TODO: allow optional kwargs
+                }
+            )
+        else:
+            print(cat.llm)
+            embedder = embedders.EmbedderAzureOpenAIConfig.get_embedder_from_config(
+                {
+                   "openai_api_key": openai_key,
+                   "model_name": "text-davinci-003",
+                   "api_base": "https://stregatto.openai.azure.com/",
```
zioproto, line 96: Can I read this value from `cat.llm.api_base` ? I was not able to figure out how to access the parameters

zioproto:
Currently blocked by https://github.com/hwchase17/langchain/issues/1560

core/cat/plugins/quickstart_cat_plugin/models.py
```diff
@@ -72,20 +72,27 @@ def get_language_model(cat):
...
         openai_key = os.getenv("OPENAI_KEY")
         if openai_key is None:
             openai_key = cat.llm.openai_api_key
-        embedder = embedders.EmbedderOpenAIConfig.get_embedder_from_config(
-            {
-                "openai_api_key": openai_key,
-                # model_name: '....'  # TODO: allow optional kwargs
-            }
-        )
+        if using_openai_llm in [OpenAI, OpenAIChat]:
```
zioproto, line 83: @pieroit should I reverse the logic here ? Check of `AzureOpenAI` and use the embedder `EmbedderAzureOpenAIConfig` and then have a catch all `else` ?

core/cat/factory/llm.py
```diff
@@ -101,6 +101,23 @@ class Config:
             "description": "Configuration for HuggingFace Endpoint language models",
         }
 
+# https://python.langchain.com/en/latest/modules/models/llms/integrations/azure_openai_example.html
+class LLMAzureOpenAIConfig(LLMSettings):
+    openai_api_key: str
+    api_base: str
+    api_type: str = "azure"
+    api_version: str = "2022-12-01"
```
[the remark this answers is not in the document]
soferreira: yes, let's keep that one for now. There is a new version already, but it is still in preview [2023-03-15-preview](https://learn.microsoft.com/en-us/azure/cognitive-services/openai/reference#completions).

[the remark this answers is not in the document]
pieroit: > model_name: davinci is fine. Can the user then override this? user may want to use cheaper/more powerful model

Yes, config classes are used to generate a json schema for the admin (web ui).
For example:
`param_name: str`
Will create an empty and required field in the admin
`param_name: str = 'value'`
Will create an editable field with `value` in it

core/cat/plugins/quickstart_cat_plugin/models.py
pieroit: @zioproto @soferreira I made some refactoring to avoid confusion between default cat behaviour and plugin extensions.

If you look into `cat/mad_hatter/core_plugin` you will find some hooks that define how the cat loads LLM and embedder.

You can override the `get_language_model` and `get_language_embedder` hooks to always have the cat loading a specific LLM/embedder. In `cat/plugins/myplugin/myfile.py`:

```python

@hook
def get_language_model(cat):

    # look here in DB, or totally skip it

    config = {
        openai_api_key: os.environ["SOME_ENV_VAR"]
        api_base: "something"
        api_type: "azure"
        api_version: "2022-12-01"
    }

    return LLMAzureOpenAIConfig.get_llm_from_config(config)

```

From now on your hook will be executed and not the default one.
Hope this is intuitive enough, feedback is welcome

zioproto:
@pieroit LGTM, feel free to merge

zioproto:
@pieroit I pushed a new commit to avoid using the Fake Embeddings. I had to upgrade LangChain to 0.0.155 to include this change: https://github.com/hwchase17/langchain/pull/3711

core/cat/mad_hatter/core_plugin/hooks/models.py
```diff
@@ -72,20 +72,37 @@ def get_language_model(cat):
 
 @hook(priority=0)
 def get_language_embedder(cat):
-    # TODO: give more example ocnfigurations
+    # TODO: give more example configurations
 
     # Embedding LLM
-    using_openai_llm = type(cat.llm) in [OpenAI, OpenAIChat]
+    using_openai_llm = type(cat.llm) in [OpenAI, OpenAIChat, AzureOpenAI]
     if ("OPENAI_KEY" in os.environ) or using_openai_llm:
```
pieroit, line 79: let's totally ignore environment variables ;)

core/cat/mad_hatter/core_plugin/hooks/models.py
```diff
@@ -72,20 +72,37 @@ def get_language_model(cat):
...
         openai_key = os.getenv("OPENAI_KEY")
         if openai_key is None:
             openai_key = cat.llm.openai_api_key
-        embedder = embedders.EmbedderOpenAIConfig.get_embedder_from_config(
-            {
-                "openai_api_key": openai_key,
-                # model_name: '....'  # TODO: allow optional kwargs
-            }
-        )
+        if using_openai_llm in [OpenAI, OpenAIChat]:
```
pieroit, line 83: `using_openai_llm` is a boolean, not a type!

aakashrshah:
Related issue https://github.com/hwchase17/langchain/issues/3992

PR #139, opened 2023-04-19
Added .env file in admin folder to set up server ip and port

Steve707-AE:


Steve707-AE (commented):
I know, I added it to upload the .env file and let people know how to use the appropriate parameters to set server ip and port..

pieroit:
@Steve707-AE ok leave as is, I'm adding a few things and merge

We can include a .env_example

pieroit:
@Steve707-AE thanks, I elaborated a little, please check.
There is now a `.env.example` with some documentation. Both containers host and port can be customized

PR #144, opened 2023-04-20
Add and improve some docstrings. Add some typing

dSupertramp:


pieroit:
Thank you for starting some inline documentation.
The methods in the `quickstart_plugin` are now moved to `cat/mad_hatter/core_plugin`, can you please update?
Sorry for the delay

dSupertramp:
Got it, I'll update all docs soon

PR #155, opened 2023-04-21
WiP: Refactor the admin container to be immutable

zioproto:
From Kubernetes docs:
https://kubernetes.io/docs/concepts/containers/

Containers are intended to be stateless and immutable: you should not change the code of a container that is already running. If you have a containerized application and want to make changes, the correct process is to build a new image that includes the change, then recreate the container to start from the updated image.

### DIFF
--- a/admin/Dockerfile
+++ b/admin/Dockerfile
@@ -1,8 +1,20 @@
-FROM node:18
+FROM node:18.16.0-bullseye-slim
 
+COPY . /app
 # Create app directory
 WORKDIR /app
+RUN adduser --disabled-password --gecos '' appuser
+RUN chown -R appuser:appuser /app
+RUN chmod 755 /app
+USER appuser
+WORKDIR /app
 
+#ENV NODE_ENV=production
+#RUN npm ci --only=production ; npm run build # Review package.json for promotion of devDependencies
+RUN npm install
 # Expose port 3000 and build + start application
 EXPOSE 3000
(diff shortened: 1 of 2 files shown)


admin/Dockerfile
```diff
@@ -1,8 +1,20 @@
...
 
+#ENV NODE_ENV=production
+#RUN npm ci --only=production ; npm run build # Review package.json for promotion of devDependencies
+RUN npm install
 # Expose port 3000 and build + start application
 EXPOSE 3000
-CMD ["/bin/bash", "-c", "npm install; npm run dev"]
+# Run npm as our non-root user
+
+#CMD ["/bin/bash", "-c", "npm run start"] # Review package.json for promotion of devDependencies
```
zioproto, line 19: @antonioru I tried to build a version of the `Dockerfile` for production to run `npm run start` but I was not able to. Do you have a working example ?

zioproto: I am fine keeping the container running in development mode, but at least let's run an immutable development container :)

I am packaging the cat for Kubernetes, but if the code is mounted at runtime I cant make it work.

please ignore the commented notes about production settings, how do you feel with the rest of the changes ? :) 

Thanks

antonioru: Lookin proper sick fam, mad respect ☺️
Appreciate your contribution

pieroit:
Too soon for this, closing the PR.
Thanks

PR #160, opened 2023-04-22
Refactored summarization chain and prompt

calebgcc:
- Removed langchain's default summarization chain 
- Added summarization prompt to core hooks


pieroit:
Beautiful thanks

PR #162, opened 2023-04-23
Added suounds toggler button

PaperBackPear3:
- created redux slice for the audio state 
- in Header added new button that if clicked stops audio from being played for both messages and microphone interactions


antonioru:
Hey @PaperBackPear3 , thank you so much for this contribution and for your commitment! 
The work you have done is amazing! 
Having a setting to disable notification sound is certainly an interesting nice-to-have, unfortunately before introducing new features in the admin they should be discussed and put into the roadmap.

There is already a settings page where in the future we could implement generic settings like this, also at the moment your implementation puts this controller on every page, but in terms of usability, it doesn't make much sense (why should there be a sound controller in the plugins page for example?).
My advice is to open a discussion about adding a series of basic settings that can be modified from the settings page.

pieroit:
@antonioru I gave him green light for this, does not look like roadmap material to me.

Also I don't think a user should click two or three times just to mute the sound.

Can you guys (@PaperBackPear3 and @antonioru ) reach agreement on how to add a simple mute button in the chat?

Thanks

PaperBackPear3:
> Hey @PaperBackPear3 , thank you so much for this contribution and for your commitment! The work you have done is amazing! Having a setting to disable notification sound is certainly an interesting nice-to-have, unfortunately before introducing new features in the admin they should be discussed and put into the roadmap.
> 
> There is already a settings page where in the future we could implement generic settings like this, also at the moment your implementation puts this controller on every page, but in terms of usability, it doesn't make much sense (why should there be a sound controller in the plugins page for example?). My advice is to open a discussion about adding a series of basic settings that can be modified from the settings page.


@antonioru  thanks for the feedback i got your point, it doesn't make much sense to have that option everywhere. it might have sense to have it only in the home section, where sounds actuvally take place? lmk

pieroit:
Thank you!

PR #168, opened 2023-04-24
Update cat-core.md, added stringified json 

salvog7:
added stringified json


EugenioPetulla:
Can you please also add an Admonition with the explicit json schema? I think the PR will be more complete. thank you 🙌🏼 

After line 17 put something like:

```markdown
!!! info "Request JSON schema"
    Sending input will request you to do it in the following specific JSON format `{"text": "input message here"}`
```

If you need some reference for admonitions, we use this: https://squidfunk.github.io/mkdocs-material/reference/admonitions/

pieroit:
Great thanks

Issue #163, opened 2023-04-23
`Docker-compose.yml`: Add prefix to service and container name


luca-rigutti:
Hi,

If there is not problem with something else on the code, I suggest to add a prefix to service and container to avoid collision with other service used with docker.

An example of this can be:

```yaml
  cheshire-cat-core:
      build:
        context: ./core
      container_name: cheshire_cat_core
```

And also for the `admin` and `vector-memory`

Thanks

pieroit:
Yes this is a good idea can you do it in a PR?

The only place I recall there is a direct reference to a container name in the code is here
https://github.com/pieroit/cheshire-cat/blob/main/core/cat/memory/vector_memory.py#L21

Please test everything with a fresh installation (I'll do the same in review)

Thanks

pieroit:
solved in #177

PR #177, opened 2023-04-26
Issue #163 docker compose naming

luca-rigutti:
Adding prefix to docker compose service to avoid collision between other services

Relative to the issue#163


pieroit:
thanks

PR #182, opened 2023-04-27
quick solution for https

TochKa21U:
Created for issue https://github.com/pieroit/cheshire-cat/issues/175


pieroit:
@Mte90 no it is not, would you like to introduce a section about environment variables in the docs?

Please not that LLM related envs are going to disappear in a few days (OPENAI_KEY, HF_ENDPOINT, ecc.)

Mte90:
So the ticket associate to this PR needs to be reopened and write it about that part.

Issue #142, opened 2023-04-20
Cat reset functionality


pieroit:
It is useful for development to have a fast way to:
- delete long_term_memory
- delete db

Both via endpoints and via cat/plugin

ian3z:
It makes sense to selectively erase long term memory?

pieroit:
@ian3z please explain yourself better

ian3z:
I would like the cat to be able to forget certain topics or files that I have entered in the past, just to eliminate some bias that can occour in current o future interactions.
So, for files ingestions, can be useful to have a lookup table where we can link the thoughts that he created linked to a certain document.
I dont know if an information like that is already present in the long memory cache.
Another useful thing is that we can make the internal prompts more lighter (using OpenAI API is expensive, if we can eliminate unuseful tokens is good)

pieroit:
> I would like the cat to be able to forget certain topics or files that I have entered in the past, just to eliminate some bias that can occour in current o future interactions.

The cat can already store memories with metdata in a vector db, and can retrieve/delete based on those metadata.
It's just not yet implemented in the cat API. There is a `/memory/{memory_id}/` endpoint to DELETE specific memories. We should add a DELETE `memory/` endpoint that receives the filters to delete memories with certain metadata.


> Another useful thing is that we can make the internal prompts more lighter (using OpenAI API is expensive, if we can eliminate unuseful tokens is good)

You can customize the prompts via plugin. This part will be improved to be more explicit

peppicus:
i have implemented a route that delete all long term memories, if it is useful it is ready for commit

pieroit:
There are endpoints to delete collections or the whole memory.
This functionality should be added to the admin in the `Memory` page.
Closing this for now as functionality is available

PR #183, opened 2023-04-28
Merge branch 'main' of https://github.com/peppicus/cheshire-cat into …

peppicus:
…empty_memories Cat reset functionality #142


peppicus:
i did two routes
/wipe_collections/ that wipe out all collections, but also history cat, and 
/collection/{collection_id} that wipe out only the collection_id, without wipe out also the history

pieroit:
Thank you @peppicus I left some comment to review a few details

Issue #188, opened 2023-04-30
color the lines in the log


peppicus:
i've ready the commit :-)

pieroit:
done in #189

PR #189, opened 2023-05-01
color the lines in the log #188

peppicus:


pieroit:
thanks this is beautiful!!!

PR #192, opened 2023-05-02
Added `WorkingMemory` class (issue #103) 

nicola-corbellini:
Added `WorkingMemory` class as proposed [here](https://github.com/pieroit/cheshire-cat/issues/103#issuecomment-1531740764)
Removed `reset_memory` from `CheshireCat` as now the `WorkingMemory` handles this


pieroit:
thanks we're going to use this a lot

PR #202, opened 2023-05-10
fix(log): no break colors

Mte90:
Now the output is:
immagine
Not this mess:
immagine

Also remove the hook_function parameter in the log https://github.com/pieroit/cheshire-cat/issues/180 as today is just the address memory doesn't give any useful information to the developer. We already know that for that plugin is loading that hook as the hook name is printed so we don't need that value.
In this way the log starts to have the stuff useful in the log.

I have added a bit of documentation to the utils.py file as I working on that.

### DIFF
--- a/core/cat/mad_hatter/decorators.py
+++ b/core/cat/mad_hatter/decorators.py
@@ -42,7 +42,6 @@ def hook(_func=None, priority=1) -> Any:
             doc_string = ""
         CatHooks.add_hook(
             {
-                "hook_function": cat_hook_wrapper,
                 "hook_name": func.__name__,
                 "docstring": func.__doc__,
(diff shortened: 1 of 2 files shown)


pieroit:
Thanks @Mte90 

Please pay attention, by removing [this line](https://github.com/pieroit/cheshire-cat/blob/main/core/cat/mad_hatter/decorators.py#L45) you broke the plugin system.

A reference to the function is necessary for the function to be actually called, that happens [here](https://github.com/pieroit/cheshire-cat/blob/main/core/cat/mad_hatter/mad_hatter.py#L118)

Mte90:
So it make sense to remove that parameter just when the log is printed.

PR #204, opened 2023-05-10
Pyproject it's here!

Mte90:
In this way we can introduce a new way in python to manage dependencies.  
We can define option dependence used only from the developers as example, from the CI and so on. It includes also the configuration for pylint/autopep8 in the same file.  
In the future can be used also for pypi to download the whole project.

Fix:

* https://github.com/pieroit/cheshire-cat/issues/178
* Now the cat version is read by the pyproject file so we are removing an hardcore reference in a file

Future:

* when is ready we can start work on https://github.com/pieroit/cheshire-cat/issues/106 to integrate something for all the codebase

It is not yet tested as I have to understand how to reset my docker to download everything again.

I have to understand also how to execute a command inside docker to download the developer stuff.


core/cat/routes/openapi.py
```diff
@@ -14,7 +15,7 @@ def custom_openapi():
 
         openapi_schema = get_openapi(
             title="😸 Cheshire-Cat API",
-            version="0.1 (beta)",
+            version=tomli.load(open("../../pyproject.toml", "rb"))["project"]["version"],
```
Mte90, line 18: I wasn't able to test this as I don't understand how to reach that code.

Mte90:
Wait for a merge as I want to integrate in this one the opportunity to download the requirements for plugins too.

Mte90:
So with pyproject we can't do it in a way that load automatically requirements.txt from plugins if there are other requirements like in our case.
At the end the result it's the same as I don't like the approach to generate a new requirements.txt file with all the dependencies of the various plugins and the cat and use it to install it.

But we can modify the Dockerfile https://github.com/pieroit/cheshire-cat/blob/main/core/Dockerfile to check every plugin folder for `requirements.txt` and install them automatically.

The pyproject.toml will be used in the core to manage the various dependences and tools to simplify the life of the contributors but the plugin will use the requirements as they will use the same dev tools of the cat so it isn't not needed right now.

pieroit:
> But we can modify the Dockerfile https://github.com/pieroit/cheshire-cat/blob/main/core/Dockerfile to check every plugin folder for requirements.txt and install them automatically.

Like it, let's do this

Mte90:
If https://github.com/pieroit/cheshire-cat/pull/233 get merged this one can be merged too.

I don't understand the conflict as this PR just remove that file...

Mte90:
Ok Fixed the issues on the PR.
I think that is ready to be merged.

pieroit:
@Mte90 thnks, there are new libs to be added to the .toml that have been added in PR #239, but I can do that myself.
Do I understand correctly that before merging this I should merge #233 ?

Mte90:
Yes in this way we do everything together but you can test the high priority PR for plugin updates.

Issue #205, opened 2023-05-10
Qdrant doesn't respond


Mte90:
In my case seems that qdrant is not responding so everything doesn't work.

In my opinion we should check if qdrant responds and avoid all this errors but generate an helpful message.
Why is not responding qdrant on my machine I have no idea but I think that is helpful to find a solution for that issue.

The full log:
```
[2023-05-10 15:31:12,063] DEBUG  _trace.py 36 -=> connection.connect_tcp.started host='cheshire_cat_vector_memory' port=6333 local_address=None timeout=5.0
[2023-05-10 15:31:12,064] DEBUG  _trace.py 36 -=> connection.connect_tcp.failed exception=ConnectError(gaierror(-2, 'Name or service not known'))
[2023-05-10 15:31:12,068] DEBUG  utils.py 67 -=> ----------------  ^._.^  ---------------- => cat.memory.py (VectorMemoryCollection.create_collection_if_not_exists(...)) @ 57 line
[2023-05-10 15:31:12,068] DEBUG  utils.py 71 -=> 'Creating collection episodic ...'
[2023-05-10 15:31:12,068] DEBUG  utils.py 72 -=> -----------------------------------------
[2023-05-10 15:31:12,068] DEBUG  _trace.py 36 -=> connection.connect_tcp.started host='cheshire_cat_vector_memory' port=6333 local_address=None timeout=5.0
[2023-05-10 15:31:12,068] DEBUG  _trace.py 36 -=> connection.connect_tcp.failed exception=ConnectError(gaierror(-2, 'Name or service not known'))
[2023-05-10 15:31:12,070] ERROR  on.py 121 -=> Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/httpcore/_exceptions.py", line 10, in map_exceptions
    yield
  File "/usr/local/lib/python3.10/site-packages/httpcore/backends/sync.py", line 94, in connect_tcp
    sock = socket.create_connection(
  File "/usr/local/lib/python3.10/socket.py", line 824, in create_connection
    for res in getaddrinfo(host, port, 0, SOCK_STREAM):
  File "/usr/local/lib/python3.10/socket.py", line 955, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno -2] Name or service not known

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #206, opened 2023-05-10
fix(qdrant): alert user when qdrant don't respond

Mte90:
The idea is to do a socket request to check if qdrant respond, otherwise block the cat execution to avoid a long traceback useless to the user.

Fix: #205

### DIFF
--- a/core/cat/memory/vector_memory.py
+++ b/core/cat/memory/vector_memory.py
@@ -1,4 +1,6 @@
 import os
+import sys
+import socket
 import time
 from typing import Any, Callable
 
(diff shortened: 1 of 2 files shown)


Mte90:
As today there is no difference between `expose` and `ports` https://stackoverflow.com/questions/65784039/why-a-service-configured-with-only-expose-is-able-to-communicate-with-internet/65785558#65785558 in dockerfile.
This change let also to get qdrant accessible from the browser so for development is more useful to check that after all is running.

Mte90:
Anyway on my tests is still not responding from the cat itself, so just wait.

Mte90:
Seems that in my case is podman with issues on dns resolution. Anyway I am working on this but the changes are still valid for other similar cases.

pieroit:
Hi @Mte90 the [official docs](https://docs.docker.com/compose/compose-file/compose-file-v3/#expose) explain:

> Expose ports without publishing them to the host machine - they’ll only be accessible to linked services. Only the internal port can be specified.

It is required that Qdrant is only visible between the docker containers but not from outside!
Otherwise anybody can just mess around with the vector db... only the core should see it.

Closing this, let me know if you have issues
Thanks

Mte90:
The pr is not just expose but also check if qdrant is running.
I can revert the port/expose stuff that is outside the real purpose of the pr.

PR #209, opened 2023-05-11
eslint bugfix, also added auto fix on pipeline

TochKa21U:
Eslint on the main has problems, it throws an error on admin-lint action ( More details: https://github.com/pieroit/cheshire-cat/actions/workflows/admin-linting.yml?page=2) 

This bugfix is targeted toward fixing current ESLint errors and warnings and adding an auto lint fix on the GitHub actions.

### DIFF
--- a/.github/workflows/admin-linting.yml
+++ b/.github/workflows/admin-linting.yml
@@ -10,6 +10,9 @@ jobs:
       - name: Install dependencies
         working-directory: ./admin
         run: npm ci
+      - name: Auto Lint Fix
+        working-directory: ./admin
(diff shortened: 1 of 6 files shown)


pieroit:
Thanks for the fix.
In the next weeks we are porting the admin to Vue + Pinia, always in typescript.
Will this setup work? Would you help us in case to accomodate the new version for the github action?

pieroit:
Thanks for the fix.
In the next weeks we are porting the admin to Vue + Pinia, always in typescript.
Will this setup work? Would you help us in case to accomodate the new version for the github action?

PR #215, opened 2023-05-13
concurrently check for new user messages and notifications

zioproto:
To achieve parallelism and concurrently check for new user messages and notifications, you can use the asyncio library in Python and convert the while loop into an asynchronous coroutine. You can then use the asyncio.gather() function to run both the message receiving and notification checking coroutines simultaneously.

### DIFF
--- a/core/cat/routes/websocket.py
+++ b/core/cat/routes/websocket.py
@@ -1,4 +1,5 @@
 import traceback
+import asyncio
 
 from fastapi import APIRouter, WebSocket
 from cat.utils import log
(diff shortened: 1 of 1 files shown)


zioproto:
@pieroit what is missing to merge this PR ?

pieroit:
thanks!
Also added a `type` key to the cat message to differentiate `chat` vs `notificatio` messages
(a client may be interested in displaying them in a different way)

zioproto:
OK, I dont see why commit 7c19951cc5c76b62b87e42e7a255995ea4b1f0a3 does not appear in this PR. But it is included in the `main` branch. All good.

pieroit:
@zioproto I guess because I merged on a local branch and then added the commit to it

PR #224, opened 2023-05-16
Create Easy-Install_Ubuntu.sh

luca1844:


pieroit:
thank you for this, but I'm not merging it.
- there are so many ways things can go differently by installing docker
- you do not run docker with sudo

Maybe later in time we can reelaborate on this script and have a command line utility, at the moment it is too soon.
Thanks anyway!

PR #225, opened 2023-05-16
Added hook `before_collection_created` to override the first point inserted in memories

nicola-corbellini:
I added the hook we discussed on the Discord.

Now `VectorMemory` expects an instance of the Cat and propagates it to the two `VectorMemoryCollections`. 
Please, let me know if it is done like you had in mind.

I followed the implementation of the [hook](https://github.com/pieroit/cheshire-cat/blob/8c7a961235cf424c48bb7f260fa8d118ea2df37a/core/cat/rabbit_hole.py#L141) you pointed out. Hence, `first_memory` is expected to be `langchain.Document`.

### DIFF
--- a/core/cat/looking_glass/cheshire_cat.py
+++ b/core/cat/looking_glass/cheshire_cat.py
@@ -89,7 +89,7 @@ class CheshireCat:
 
     def load_memory(self):
         # Memory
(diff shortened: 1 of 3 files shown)


pieroit:
Perfect thanks

PR #231, opened 2023-05-20
Add hooks for translating documents / url

cristianorevil:
As discussed on the discord channel it would be useful to have 2 hooks in the rabbit hole to allow external plugins to manipulate (eg translation) the content of the document or url.

The proposed hooks are:

before_rabbithole_splits_text(doc, cat) to be executed as soon as the document is loaded and before being split

after_rabbithole_splitted_text(chunks, cat) to execute after splitting the document into chunks

I have tested this hook with a multilingual plugin (currently in develop) and i have succefully translated document and url (from italian to english)


pieroit:
thanks!!

PR #232, opened 2023-05-20
Add hook after cat bootstrap

cristianorevil:
Adding an hook after cat bootstrap can be useful to dev for access cat instances (eg adding a value in working memory) when the cat bootstrap without waiting user interaction

### DIFF
--- a/core/cat/looking_glass/cheshire_cat.py
+++ b/core/cat/looking_glass/cheshire_cat.py
@@ -42,6 +42,8 @@ class CheshireCat:
         # Rabbit Hole Instance
         self.rabbit_hole = RabbitHole(self)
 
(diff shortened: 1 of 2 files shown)


pieroit:
@cristianorevil  thanks for adding the hook! It's going to be really useful.

I changed the name because I realized we can have one hook before bootstrap and another after, so there are two hooks: `before_cat_bootstrap` and `after_cat_bootstrap`.

From now on let's follow this nice convention on cat hooks *naming* proposed by @EugenioPetulla :
 
 - `WHEN_WHO_WHAT`
 - `before_cat_bootstrap`
 - `before_rabbithole_insert_memory`

Checking out your other PR in a minute

PR #233, opened 2023-05-20
install plugins requirements

### LINKED ISSUE #203: Docker automatically installs requirements.txt files from plugins
At the moment the docker image is built by running `pip install --no-cache-dir -r requirements.txt` in `core/Dockerfile`.
The dockerfile should also search in the `core/cat/plugin` folder and install any `requirements.txt` found in there.

That way an installation can be updated with the latest version of the cat, respecting plugin dependencies.

alexcalabrese:
## Related Issue
Fixes #203.

## Changes proposed in this pull request
- Created a bash script that searches and installs all the "requirements.txt" inside the plugins folder.
- Updated Dockerfile to run that script.

### DIFF
--- a/core/install_plugin_dependencies.sh
+++ b/core/install_plugin_dependencies.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+find ./cat/plugins -type f -name "requirements.txt" | while read -r req_file; do
+    pip install --no-cache-dir -r "$req_file"
+done
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -7,3 +7,9 @@ RUN apt-get update -y
 COPY ./requirements.txt requirements.txt
 RUN pip install -U pip
 RUN pip install --no-cache-dir -r requirements.txt
+
+COPY ./cat/plugins ./cat/plugins
+COPY ./install_plugin_dependencies.sh install_plugin_dependencies.sh
+
+RUN chmod +x ./install_plugin_dependencies.sh
+RUN ./install_plugin_dependencies.sh


core/install_plugin_dependencies.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+find ./cat/plugins -type f -name "requirements.txt" | while read -r req_file; do
+    pip install --no-cache-dir -r "$req_file"
+done
```
Cereal84, line 5: Hi,
I'm totally new to this repo in which I have no experience and this is just a comment not a review.
Anyway, I thought that this way to handles dependencies could be troublesome. 
Think about having different plugins which use the same library but support a different version, based on the order in which the scripts install the dependencies one or more script cannot be run correctly.
Probably it needs a script, bash or python or something else, which reads all the dependencies and their version and creates a new requirements.txt with the list of libraries and the minimum version or return an error if the versions of the same library are not compatible (jinja<2.0, jinja>=3.X).

These are my 2 cents about this PR.

pieroit:
Thank you @alexcalabrese.
I agree with @Cereal84 the solution to plugin dependencies is going to be more elaborate, maybe this is a start.

What's your take here @Mte90 ?

pieroit:
@Cereal84 thanks for your insight.
The plan could be:
- have this simple solution as a starting point
- merge also .toml #204 to better manage main dependencies
- improve this script to avoid plugin dependencies conflicts

pieroit:
We got a problem on windows with file endings (LF vs CRLF), see #253

PR #235, opened 2023-05-22
Workaround to speed up pdf loading time

zioproto:
Related to:
https://github.com/pdfminer/pdfminer.six/issues/347

pdfminer with log set to DEBUG is super slow because it prints way too many debug statements.


pieroit:
Thanks for speeding this up!

PR #237, opened 2023-05-22
Pass the text_splitter to the loader

zioproto:
The `text_splitter` in LangChain is supposed to be passed to the as a parameter to `load_and_split()`

### DIFF
--- a/core/cat/rabbit_hole.py
+++ b/core/cat/rabbit_hole.py
@@ -92,17 +92,17 @@ class RabbitHole:
         :param chunk_overlap: number of overlapping characters between consecutive chunks
         """
 
-        # load text content of the website
-        loader = UnstructuredURLLoader(urls=[url])
-        text = loader.load()
-
-        # split in documets using chunk_size and chunk_overlap
+        # Initialize the text splitter with custom separators chunk_size and chunk_overlap
         text_splitter = RecursiveCharacterTextSplitter(
             chunk_size=chunk_size,
             chunk_overlap=chunk_overlap,
             separators=["\\n\\n", "\n\n", ".\\n", ".\n", "\\n", "\n", " ", ""],
         )
-        docs = text_splitter.split_documents(text)
+
+        # load text content of the website
+        loader = UnstructuredURLLoader(urls=[url])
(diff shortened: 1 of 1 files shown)


zioproto:
Why these default values have been changed ?

Default values:
https://github.com/hwchase17/langchain/blob/ef7d015be566588b3263ee6ee1259a30ee53212c/langchain/text_splitter.py#L33-L36
https://github.com/hwchase17/langchain/blob/ef7d015be566588b3263ee6ee1259a30ee53212c/langchain/text_splitter.py#L275-L278

Our values:
https://github.com/pieroit/cheshire-cat/blob/main/core/cat/rabbit_hole.py#L103
https://github.com/pieroit/cheshire-cat/blob/main/core/cat/rabbit_hole.py#L26-L27

Why our values are better than the LangChain default values ?

@calebgcc

calebgcc:
Hi @zioproto, I think @pieroit came up with the values by experimenting with different documents, but I am not sure.

The problem I think these values solve is that when we summarise we merge 3 documents (3 is the default)
We cannot use long texts as input for the models because some of them have a strict maximum length.

pieroit:
@zioproto see here the implementation of `split_and_load`
https://github.com/hwchase17/langchain/blob/master/langchain/document_loaders/base.py#L27

By doing the same thing out of langchain, we can insert there hooks that may be useful to people.
For example you may want to do something with the loaded content before it is splitted.

For the same reason we didn't use llama-index and many other utilities.
Please think of the Cat as a framework, not as a product.
Thanks :)

zioproto:
> separators=["\\n\\n", "\n\n", ".\\n", ".\n", "\\n", "\n", " ", ""],

What about this list of separators ?

zioproto:
> > The problem I think these values solve is that when we summarise we merge 3 documents (3 is the default) We cannot use long texts as input for the models because some of them have a strict maximum length.
> 
> Can confirm, with 3 docs of length 4000 there were many "out of context" errors. @zioproto which defaults do you suggest?

If I understand correctly, the problem is with the token limit of the Embedding API.

If you know the number of maximum tokens the API accepts, you can use it to tune `chunk_size` depending on the configured embedding model

Issue #239, opened 2023-05-22
Add support for PaLM and Claude


pieroit:
Langchain supports both [PaLM by Google](https://python.langchain.com/en/latest/reference/modules/llms.html#langchain.llms.GooglePalm) and [Claude by Anthropic](https://python.langchain.com/en/latest/reference/modules/llms.html#langchain.llms.Anthropic).

These can be added as LLM to the Cat by updating the [LLM factory](https://github.com/pieroit/cheshire-cat/blob/main/core/cat/factory/llm.py), in a similar way as other language models are supported via langchain.

PR #240, opened 2023-05-22
Add support for PaLM and Claude #239 

Zap123:
This pull request resolves _add support for PaLM and Claude #239_. 
langchain.llms.Anthropic was marked as deprecated in the langchain class [here](https://python.langchain.com/en/latest/_modules/langchain/llms/anthropic.html) so I used ChatAnthropic instead.
I've also added to the dependencies "anthropic" and "google-generative" ai needed for these new LLMs to work.
I currently don't have access to an api key for those models but I've verified that after setting the model as default I get "401 unauthorized" error if I try to send a new message.

### DIFF
--- a/core/cat/factory/llm.py
+++ b/core/cat/factory/llm.py
@@ -60,6 +60,28 @@ class LLMOpenAIChatConfig(LLMSettings):
             "description": "Chat model from OpenAI",
         }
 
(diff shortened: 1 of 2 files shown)


pieroit:
Thanks,!
I'll take the respective API keys and try them out, then review and merge.

PR #241, opened 2023-05-22
Inline hooks documentation

nicola-corbellini:
- module-level documentation in `flow.py`
- documented `before_cat_reads_message`
- documented `before_cat_recalls_memories`


pieroit:
Thanks

PR #244, opened 2023-05-25
New log engine

Mte90:
Ref: https://github.com/pieroit/cheshire-cat/issues/43

This new log engine implements:

* 2 new environment variables
  * LOG_LEVEL, set the global level to show for everything (default in code DEBUG)
  * LOG_LEVEL_DEPENDENCIES, se the global level to use for the python dependencies (default in code DEBUG)
* Native `loguru` system for log colors (no broken output anymore)
* Removes the verbose parameters in various glasses as now the `log` function have a parameter where you can specify the kind of verbosity
* In DEBUG level shows for any call the stacktrace (only for log request from the cat) in a red color 
* Improve the output for JSON in terminal

This means that if we set DEBUG it will be showed any log from the cat, if we set INFO we see only the log request configured as INFO.
With the other variable we change the behavior for the dependencies, as there is a code that change the log level for all the dependencies (example SQLalchemy) to this new variable.
With that code (that was there before), if the user set as example for LOG_LEVEL_DEPENDENCIES as DEBUG and the cat as INFO it will never see the log stuff from sqlalchemy or the other libraries.

So the new approach is different and requires less code to maintain, as in every log request we specify the log levels. 
Those log levels have different colors, also it is possible to set to save any of them in a different file automatically (by the loguru library itself). 
It is possible to create custom log levels (for the future): https://loguru.readthedocs.io/en/stable/overview.html#customizable-levels
The standard levels are: 

        +----------------------+------------------------+------------------------+
        | Level name           | Severity value         | 
        +======================+========================+========================+
        | ``TRACE``            | 5                      | 
        +----------------------+------------------------+------------------------+
        | ``DEBUG``            | 10                     | 
        +----------------------+------------------------+------------------------+
        | ``INFO``             | 20                     | 
        +----------------------+------------------------+------------------------+
        | ``SUCCESS``          | 25                     | 
        +----------------------+------------------------+------------------------+
        | ``WARNING``          | 30                     |
        +----------------------+------------------------+------------------------+
        | ``ERROR``            | 40                     | 
        +----------------------+------------------------+------------------------+
        | ``CRITICAL``         | 50                     | 

This is the priority order used also to show them, so if the user set CRITICAL it should see all of them and so on.
We can improve the log engine to change the behaviour to show only the same level configured as example.

This is a DEBUG example, if it is configured INFO in both there is only the info stuff that it is few things like plugin loaded (without the list) etc.

immagine

This is the new output for plugins and hooks:

immagine

Before closing the PR I suggest to try it and do some evaluations about the log levels we want to set and where.

A suggestion I can say is to set as default for LOG_LEVEL_DEPENDENCIES as critical, in this way they are not showed if LOG_LEVEL is DEBUG. In this way it is more simple for everyone to have only the output from the cat and not from qdrant, sqlalchemy etc.
Another one is to see to change the DEBUG log level as severity 45 in this way we get also warning and errors (probably we can do some tests).


pieroit:
Thanks this is great.

I do not understand the reason for `LOG_LEVEL_DEPENDENCIES` and how it is related to `LOG_LEVEL`

> A suggestion I can say is to set as default for LOG_LEVEL_DEPENDENCIES as critical, in this way they are not showed if LOG_LEVEL is DEBUG. In this way it is more simple for everyone to have only the output from the cat and not from qdrant, sqlalchemy etc.

??

pieroit:
@Mte90 merged, seems good to me.
Could you update the docs with an explanation on how this log levels work?
The difference between `LOG_LEVEL_DEPENDENCIES` and `LOG_LEVEL` is not clear to me

Mte90:
* LOG_LEVEL_DEPENDENCIES set the log level for sqlalchemy, qdrant and the other library that the project uses and that use `logging` library internally
* LOG_LEVEL it is the log level for the cat stuff but also globally that let to hide other log level also for the dependencies

In this way if you are working on the cat you can set both to debug, if you are just hosting you can set the cat level to info and the libraries logging are hidden.

Issue #47, opened 2023-03-24
Create new VectorStore collections whenever the embedder changes


pieroit:
If there is an embedder change, the VectorStore will not be compatible because:
- the new embedder may have a different output dimensionality
- even if the dimensionality is the same, it is a totally different space

So whenever the embedder changes:
- delete the old VectorStore collections
- create new ones (this is already done lazily and works perfectly if there are no VectorStore collections on disk)

Or if we want to preserve old memories, prepend embedder name to collection mane

pieroit:
Partial fix: the cat itself advices user in the chat about a change in dimensionality for the embedder and invites the user to delete the `web/long_term_memory` folder.
After reintroducing Qdrant we can find a more elegant solution.
Leaving the issue open

nickprock:
I coul work on this issue

pieroit:
@nickprock change of embedder at runtime is at the moment deactivated 

To reactivate decomment here:
https://github.com/pieroit/cheshire-cat/blob/main/core/cat/routes/setting/embedder_setting.py#L54

Also to change embedder from the admin reactivate this button (take away the `disabled` attribute):
https://github.com/pieroit/cheshire-cat/blob/main/admin/src/views/SettingsView.vue#L44

You should see that a change of embedder resulta in an error (or the cat says that embedders have different dimensionality

Let me know if this is clear enough, otherwise I'll give more references

nickprock:
Thanks @pieroit , I was able to recreate the error it was giving me before the changes. 

```
{"error":"Wrong input: Vector inserting error: expected dim: 1536, got 4096"}
```

Starting tonight I will get on it, I already have some ideas.

pieroit:
> Thanks @pieroit , I was able to recreate the error it was giving me before the changes.
> 
> ```
> {"error":"Wrong input: Vector inserting error: expected dim: 1536, got 4096"}
> ```

Great!

> 
> Starting tonight I will get on it, I already have some ideas.

I'm glad for your help. Let me know how can we fix this!
The easiest path I see is just wipeout the old collections and recreate them (that is done automagically when you bootstrap the cat). As a solution is easy but a little violent - all memories get lost for an embedder change.
Dunno

nickprock:
Hi @pieroit , 
I tried to split the problem:
1. configure the embedding dimension dinamically
2. create new VectorStore collections whenever the embedder changes

At the moment, for point 1, if in `vector_memory.py` I print the embedder:

```
[...]

PR #249, opened 2023-05-27
Dev/issue 47

nickprock:
related on issue #47 

Rough edit to not explode on embedder change.
I am looking for a more elegant solution don't close the issue please.


nickprock:
@pieroit I merged with the last version of the cat and added vector quantization for efficient storage and search. There will need to have users test it to see if the speed and quality of responses is good.

PR #250, opened 2023-05-27
Doc dockercompose

danieleongari:
Addressing (small) issue: https://github.com/pieroit/cheshire-cat/issues/248

I just updated the README because I don't know where the source of the documentation (https://pieroit.github.io/cheshire-cat/) is!


pieroit:
Thanks!
The docs (and instruction for them) are under the `docs` branch

PR #252, opened 2023-05-28
Feature/test setup

rafleze:
This PR is preparing the repo to have automatic tests.
A new file test.in is added. It contains all the requirements needed to run tests and coverage. It also add some linter.
The command `python3 -m black .`, `python3 -m ruff --fix .`, `python3 -m mypy --no-site-packages .` has been executed and some files are reformatted.
The first test is implemented.
The command `make test` is added to run all implemented tests.


core/Dockerfile
```diff
@@ -8,7 +8,7 @@ RUN apt-get update -y
 
 #RUN apt-get install -y poppler-utils tesseract-ocr libreoffice ### needed form unstructured?
 
-COPY ./requirements.txt requirements.txt
+COPY ./requirements/base.txt requirements.txt
```
Mte90, line 11: The project is not using pyproject.toml so the requirements need to be moved in that file also for the dev stuff

Mte90:
As it is the PR is not good to get an approval.
We have a pyproject.toml file now for the dependencies, that support also dev stuff and introduce pylint (also if not yet used).
I suggest to revert the various linting changes and just focus on tests and modify the pyproject file.

core/Dockerfile
rafleze: Actually I use pyproject.toml running dis command: https://github.com/pieroit/cheshire-cat/pull/252/files#diff-d486e601a844f08eaa052d963bedbdd8a7db76d664a5f44fe34bf4fd6db1064aR4

The plan is to automate a pipeline to check the code and to run tests.
To separate requirements needed by tests from projects dependencies I use a directory collecting all the requirements files.

Mte90: Sure, pyproject support dev list for requirements. In my PR waiting for approval there is already. In this way there is only just one file and with different commands you install what you want.

https://github.com/pieroit/cheshire-cat/pull/204

pieroit:
@rafleze thank you for this PR!

I'll ask you to please:
- fix conflicts (I merged PR #204 with .toml setup)
- take away linter edits, we'll see to the linter later
- core contributors (just asked) want pytest instead of unittest
- small PRs please :)

Be patient because this is delicate. Thanks again

rafleze:
@Mte90 @pieroit thanks for your feedback. The next PR will be closer to your guidelines.

PR #254, opened 2023-05-28
replace sh with python script

### LINKED ISSUE #253: Replace `install_plugin_dependencies.sh` with a python script
On windows a problem with file endings is giving errors like this:

image-1

We should get rid of .sh scripts and do everything with a python script (if no better solution is found)

This also opens up possibilities to better manage plugin dependencies as discussed in #233 

@alexcalabrese if you are interested in doing it comment and I'll assign the issue to you

alexcalabrese:
## Related Issue
Fixes #253 .

## Changes proposed in this pull request
- Replaced bash with a python script.
- Updated Dockerfile to run that script.

### DIFF
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -13,10 +13,10 @@ RUN pip install -U pip
 RUN pip install --no-cache-dir -r requirements.txt
 
 COPY ./cat/plugins ./cat/plugins
-COPY ./install_plugin_dependencies.sh install_plugin_dependencies.sh
+COPY ./install_plugin_dependencies.sh install_plugin_dependencies.py
 
-RUN chmod +x ./install_plugin_dependencies.sh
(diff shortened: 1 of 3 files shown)


pieroit:
Thanks for speed and quality!

core/Dockerfile
```diff
@@ -13,10 +13,10 @@ RUN pip install -U pip
 RUN pip install --no-cache-dir -r requirements.txt
 
 COPY ./cat/plugins ./cat/plugins
-COPY ./install_plugin_dependencies.sh install_plugin_dependencies.sh
```
pieroit, line 16: corrected to 
`COPY ./install_plugin_dependencies.py install_plugin_dependencies.py`

pieroit:
Thanks, merged!

PR #259, opened 2023-05-30
Dockerfile improvements

zioproto:
@pieroit this PR is proposed with 2 individual commits to give you the flexibility to review them independently.
You could decide to merge just 1 commit if changing the base image is a too big change.

Why we need this PR:

* Install `nodejs` manually to avoid deb package dependencies like `python-minimal:3.9`
* Change base docker image to `python:3.10.11-slim-bullseye` to reduce size, reduces image size to less than 1Gb (half the previous size)
* Makes possible to have an image that runs on at least `amd64` and `arm64` as native architectures

### DIFF
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -1,21 +1,35 @@
 
-FROM tiangolo/uvicorn-gunicorn:python3.10
+FROM python:3.10.11-slim-bullseye
 
 ### CORE ###
 
-COPY ./pyproject.toml pyproject.toml
-COPY ./cat/plugins ./cat/plugins
-COPY ./install_plugin_dependencies.py install_plugin_dependencies.py
+RUN mkdir -p /app && mkdir -p /admin
+COPY ./pyproject.toml /app/pyproject.toml
+COPY ./cat/plugins /app/cat/plugins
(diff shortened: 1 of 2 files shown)


core/Dockerfile
```diff
@@ -1,36 +1,40 @@
...
+      arm64) NODEARCH='arm64';; \
+      armhf) NODEARCH='armv7l';; \
+      i386) NODEARCH='x86';; \
+      *) echo "unsupported architecture"; exit 1 ;;\
+    esac &&\
+    pip install -U pip && \
+    pip install --no-cache-dir /app/ &&\
+    python3 /app/install_plugin_dependencies.py &&\
 ### ADMIN (static build) ###
+    curl -sL https://github.com/pieroit/cheshire-cat-admin/archive/main.tar.gz | tar xz -C /admin --strip-components=1 &&\
```
Mte90, line 29: this for sure is more fast then use git to download it

pieroit:
Thank you @zioproto it works and the image is lighter.
I reorganized a litlle for readablility and installed `build-essentials` because many python packages that may be requested from plugins require a compiler.

PR #260, opened 2023-05-30
Add link status check and error handling in the API endpoint

0siride:
implements a link status check and error handling directly in the body of the API endpoint. The code sends a HEAD request to the provided URL and captures the response status code. If the status code is 200, it assumes the URL is valid and proceeds to perform additional operations. However, if the status code is not 200, it returns a response indicating an invalid URL. Additionally, an exception block is included to handle any RequestException that may occur during the request process. In such cases, the response informs that the link is unreachable.


pieroit:
Thanks!

PR #261, opened 2023-05-31
GitHub Action CI

Mte90:
Ref: https://github.com/pieroit/cheshire-cat/issues/106, #204


Mte90:
I suggest to do a squash commit on merging in this way it is a unique commit.

Anyway this PR implement pylint with some rules and also perflint. In the next days with the other PR we can see to configure rules to excludes.

The CI it will be executed to any PR.

PR #266, opened 2023-06-01
Static media server

nicola-corbellini:
Fix for [this issue](https://github.com/pieroit/cheshire-cat/issues/218).

The file is correctly served at `http://localhost:1865/static/sample_img.png`.

However, when using a tool it's not displayed in the chat.
immagine

I tried both `f"[img](http://localhost:1865/static/{tool_input})"` and `f"[img](../static/{tool_input})"` as tool outputs, but I only get that alternative text link (which works fine).


pieroit:
@nicola-corbellini great thanks!
How do you save files into the static folder from your plugin?

pieroit:
@nicola-corbellini it works fine!

In the future we may consider having easy methods to store and get files from the static folder from within plugins

Issue #172, opened 2023-04-26
Start a new chat&conversation in Admin GUI


KevinZhang19870314:
It would be wonderful that we can have a button to start a new chat in GUI.
I found that even I refresh the http://localhost:3000 page, the `chat_history` is still in current conversation, but I just want a new chat, I don't need previous chat context.

a workround is to `docker-compose down` and `docker-compose up`, then we can have a fresh chat, but obviously it is not really good method to start a new chat yet.

pieroit:
@KevinZhang19870314 this is a good idea and must be implemented, as soon as we reach agreement on #62 
Hope you can read there and leave your view on how to proceed

cristianorevil:
Please assign to me

pieroit:
Let's decompose this in two steps:

1. endpoint DELETE `/memory/working-memory/conversation-history`
2. add reset conversation in admin

zAlweNy26:
@pieroit should the reset button be in the chat view or in the settings?

pieroit:
@zAlweNy26 let's put the button in the chat toolbar as this is much requested

Before coding it let's redesign the chat wireframe a little

PR #273, opened 2023-06-03
ADD Delete conversation history from working memory endpoint

cristianorevil:
This PR add a new endpoint /working-memory/conversation-history that DELETE the conversation history from working memory

Also change the endpoint url /wipe_collections/ to /wipe-collections/ to standardize the endpoints with the dash as discussed in public call on discord channel.

This PR is related to this issue: https://github.com/cheshire-cat-ai/core/issues/172 and solve step 1 required by @pieroit

### DIFF
--- a/core/cat/routes/memory.py
+++ b/core/cat/routes/memory.py
@@ -55,7 +55,7 @@ async def collection(request: Request, collection_id: str = "") -> Dict:
 
 
 # DELETE all collections
-@router.delete("/wipe_collections/")
+@router.delete("/wipe-collections/")
 async def wipe_collections(
     request: Request,
 ) -> Dict:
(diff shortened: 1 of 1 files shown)


core/cat/routes/memory.py
```diff
@@ -73,3 +73,17 @@ async def wipe_collections(
     ccat.load_memory()  # recreate the long term memories
 
     return to_return
+
+#DELETE conversation history from working memory
+@router.delete("/working-memory/conversation-history")
```
zAlweNy26, line 78: I think you should add a / in the end

cristianorevil: Yes, my fault. Just added

pieroit:
thanks @cristianorevil for the addition and thanks @zAlweNy26 for reviewing

this is useful also to have a "reset convo" button both in the chat widget and in the memory page

PR #290, opened 2023-06-11
Closes #289 (using node+python image in Dockerfile)

### LINKED ISSUE #289: Dockerfile simplification
The project's [Dockerfile](https://github.com/cheshire-cat-ai/core/blob/main/core/Dockerfile) now manually installs Node.js, we can instead use a docker image that already contains python and node to simplify it.

lucaGazzola:
minor change to use docker image already containing node instead of installing it.

### DIFF
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -1,5 +1,5 @@
 
-FROM python:3.10.11-slim-bullseye
+FROM nikolaik/python-nodejs:python3.11-nodejs20
 
 ### PREPARE BUILD WITH NECESSARY FILES AND FOLDERS ###
 RUN mkdir -p /app && mkdir -p /admin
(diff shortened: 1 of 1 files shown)


zAlweNy26:
I think we should stick with nodejs v18, v20 is pretty new and may cause issues.

zioproto:
The image `nikolaik/python-nodejs:python3.10-nodejs18-slim` has a problem.

You would expected python3.10, but because it is installing nodejs with deb packages, nodejs pull in the image also python3.9 as a dependency. So you have python installed 2 times in the container.

```
% docker run --rm -ti nikolaik/python-nodejs:python3.10-nodejs18-slim /bin/bash
root@2628a5760f14:/# dpkg -l | grep python
ii  libpython3.9-minimal:arm64 3.9.2-1                      arm64        Minimal subset of the Python language (version 3.9)
ii  libpython3.9-stdlib:arm64  3.9.2-1                      arm64        Interactive high-level object-oriented language (standard library, version 3.9)
ii  python3-minimal            3.9.2-3                      arm64        minimal subset of the Python language (default python3 version)
ii  python3.9                  3.9.2-1                      arm64        Interactive high-level object-oriented language (version 3.9)
ii  python3.9-minimal          3.9.2-1                      arm64        Minimal subset of the Python language (version 3.9)
root@2628a5760f14:/#
```

Tagging the author @nikolaik . @lucaGazzola maybe you want to open an issue at https://github.com/nikolaik/docker-python-nodejs.

pieroit:
@lucaGazzola thanks for this and for accomodating @zAlweNy26 request (to which I agree)

Agree also with @zioproto's observation, we should have only one python version.
At the moment node is necessary because we are building the admin from source code, in the future the admin repo should have a release so we can download a built version of the admin directly - at that point node is not necessary at all.

The most important thing right now is that the docker does not explode for RAM requirements - which occurs for most people who encountered the bug during the [curl command in dockerfile](https://github.com/cheshire-cat-ai/core/blob/main/core/Dockerfile#L33).

So it's either we make the build less heavy on ram usage or we include ram requirements in the docker-compose.yml

Let's leve this PR open for a bit, give me yours

lucaGazzola:
I am not sure you can include ram/cpu requirements for the build in the docker-compose.yml (usually those are the deploy request/limit for the services), but I am maybe mistaken. I agree that the image is not really "slim", I'll try to find a solution to that.

nikolaik:
> You would expected python3.10, but because it is installing nodejs with deb packages, nodejs pull in the image also python3.9 as a dependency. So you have python installed 2 times in the container

Thanks for mentioning me. I'll update you with my findings in https://github.com/nikolaik/docker-python-nodejs/issues/85

Issue #292, opened 2023-06-12
[BUG] Loop after uploading pdf with only images


zAlweNy26:
**Describe the bug**
I tried to upload a pdf with only images and it started looping over
`cheshire_cat_core           | Building summaries over 0 chunks. Please wait.`
without end.

**To Reproduce**
Steps to reproduce the behavior (as an example):
1. Go to the chat
2. Upload a pdf with only images
3. See the logs

**Expected behavior**
I expect the cat to respond with an information notification saying that it has created 0 thoughts on that file, otherwise, an error notification saying that it was unable to create thoughts on that file.

tarunsamanta2k20:
I'd like to look into this.

pieroit:
@tarunsamanta2k20 thanks

Tell me if you need indications or explanations on how the code works

PR #299, opened 2023-06-15
Fix loop on empty file upload

lorenzopalloni:
## Description

This pull request fixes the bug described in issue #292. The issue occurred when uploading a PDF file containing only images, resulting in an endless loop during summary creation. This pull request addresses the issue by adding a check for empty documents and returning an empty list when no documents are present.

Additionally, I would be happy to discuss [this line of code](https://github.com/cheshire-cat-ai/core/blob/main/core/cat/mad_hatter/core_plugin/hooks/rabbithole.py#L93). I would remove it to allow even short documents to be processed by the summarizer hook. Would you have any thoughts to share about it?

## Type of Change

## Checklist

### DIFF
--- a/core/cat/mad_hatter/core_plugin/hooks/rabbithole.py
+++ b/core/cat/mad_hatter/core_plugin/hooks/rabbithole.py
@@ -135,9 +135,13 @@ def rabbithole_summarizes_documents(docs: List[Document], cat) -> List[Document]
         cat: Cheshire Cat instance.
 
     Returns:
-        list of langchain`Document` with text summaries of the original ones.
+        list of langchain `Document` with text summaries of the original ones.
 
     """
+
+    if not docs:
+        return []
+
     # service variable to store intermediate results
     intermediate_summaries = docs
 


pieroit:
@lorenzopalloni thanks!

If we remove that line there will be docs and embeddings related to texts shorter than 10 characters, in which cases do you see that useful?

lorenzopalloni:
Maybe I'm wrong, but I believe there are chances to miss relevant information, while redundancy can be easily summarized. For example, if we try to split `alice.txt`, using the `langchain.text_splitter.RecursiveCharacterTextSplitter`, using the cheshire-cat's default values of `chunk_size` and `chunk_overlap`, respectively 400 and 100, I obtain that the only document filtered out would be "THE END".

```python
from langchain.text_splitter import RecursiveCharacterTextSplitter
from langchain.docstore.document import Document

chunk_size = 400
chunk_overlap = 100

text_splitter = RecursiveCharacterTextSplitter(
    chunk_size=chunk_size,
    chunk_overlap=chunk_overlap,
    separators=["\\n\\n", "\n\n", ".\\n", ".\n", "\\n", "\n", " ", ""],
)
with open("alice.txt", "r") as f:
    text = f.read()
splitted_text = text_splitter.split_text(text)
filtered_splitted_text = [t for t in splitted_text if len(t) > 10]
print(f"{len(text)=}\n{len(splitted_text)=}\n{len(filtered_splitted_text)=}")
print([t for t in splitted_text if len(t) <= 10])
```

Output:
```
len(text)=148540
len(splitted_text)=524
len(filtered_splitted_text)=523
['THE END']
```

PR #305, opened 2023-06-19
vector spaces with alias

nickprock:
# Description

Added a control on embeddings - vector_store using qdrant alias. Solves the problem of different embedders but with the same size.

Related to issue #47 

## Type of change

# Checklist:


nickprock:
Added oversampling (commented). From the qdrand documentation:

> oversampling - Defines how many extra vectors should be pre-selected using quantized index, and then re-scored using original vectors. For example, if oversampling is 2.4 and limit is 100, then 240 vectors will be pre-selected using quantized index, and then top-100 will be returned after re-scoring. Oversampling is useful if you want to tune the tradeoff between search speed and search quality in the query time.

pieroit:
thanks for your contribution on vector memory! :D

PR #307, opened 2023-06-20
Add memory threshold to filter items

cristianorevil:
# Description

This PR add a threshold value to the hook after_cat_recalled_memories to "filter" memory items. The items that not have a "score" (cosine similarity) equal or grater are removed from episodic and decalrative memory to have a context in the prompt "closest" to the user input.
The default value of threshold is 0.8 because after multiple test is the best value i have found. The threshold can be overrided in a hook with highest priority.

Related to issue #(issue)

## Type of change

# Checklist:


pieroit:
Thanks.
I found out there is a parameter in Qdrant search called [`score_threshold`](https://qdrant.tech/documentation/concepts/search/#filtering-results-by-score).

I tried it out [here](https://github.com/cheshire-cat-ai/core/blob/main/core/cat/memory/vector_memory.py#L143) but seems a little buggy (and openai api today is really slow).
Let's check it out in the future! At the moment this hook does the job perfectly ;)

PR #319, opened 2023-06-24
#275 fixed the route /static, now it is procted if the API_KEY is set

### LINKED ISSUE #275: Static files in `/static` route are not protected from `API_KEY`
**Describe the bug**

`/static` serves files even if there is an active `API_KEY` 

**To Reproduce**
1. Create `.env` in root folder
2. insert `API_KEY=meow` in the `.env`
3. put a file `a.txt` in the `core/cat/static` folder
4. open the cat at `/static/a.txt`

**Expected behavior**
API_KEY should block all endpoints

**Additional context**
Let's look again at this once #169 is fixed

giovannialbero1992:
# Description

Fix #275 

Related to issue #(issue)

## Type of change

# Checklist:

### DIFF
--- a/core/cat/routes/static/auth_static.py
+++ b/core/cat/routes/static/auth_static.py
@@ -1,0 +1,12 @@
+from fastapi.staticfiles import StaticFiles
+from fastapi import Request
+from cat.api_auth import check_api_key
+
+class AuthStatic(StaticFiles):
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+
+    async def __call__(self, scope, receive, send) -> None:
+        reqeust = Request(scope, receive=receive)
+        check_api_key(reqeust.headers.get("access_token"))
+        await super().__call__(scope, receive, send)
--- a/core/cat/routes/static/static.py
+++ b/core/cat/routes/static/static.py
@@ -1,5 +1,5 @@
-from fastapi.staticfiles import StaticFiles
+from cat.routes.static.auth_static import AuthStatic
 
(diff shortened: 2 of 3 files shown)


giovannialbero1992:
@pieroit I added a test to cover the scenario without an API_KEY set.
I'd like to add also the scenario with an API_KEY but we should understand how to manage the different app instance to manipulate any environment variable based on the test that we would perform.

pieroit:
@giovannialbero1992 thanks, added also a test for a specific file.

About the .env, you are right but I don't know how to change env variables at runtime from the test suite. Let's ask @lorenzo-polaris or anybody who wants to chip in.

lorenzo-polaris:
If you want to change env variables (or any kind of variable) at runtime, you can monkeypatch it. For instance, in your test case:

`def test_call_with_api_key(client, monkeypatch):
    # prepare
    monkeypatch.setenv("API_KEY", "test_api_key")

    # act
    response = client.get("/static/")

    # assert
    assert response.status_code == 200`

giovannialbero1992:
Hi @lorenzo-polaris, I tried your proposal but I'm not observing the tests fail. I think it's due to the fact that the application is already running, and we can't change during the runtime the environment's variable.

lorenzo-polaris:
@giovannialbero1992 tests run in a separate (clean) environment that is created and deleted each time you run pytest.

The issue you are facing is probably related to tests that use websockets, they are still connecting to the main db. It’s a bug discussed with @pieroit last week and I will fix it asap, hopefully by tomorrow!

If you are facing this problem with tests that don’t use websockets, please share the code and I will look into it!

PR #331, opened 2023-07-03
Support from new prompt settings

cristianorevil:
# Description

This PR add core support for 2 new prompt settings available in admin GUI:

Prompt prefix
Use procedural memory (=tools)
I have also introduced a regular expression that purge propmt when there are multiples empty new lines, this permit to have short prompt and read better the log.

To do this PR works i have changed in cheshire_cat.py line 151 the default prompt prefix value from:

"prefix": self.mad_hatter.execute_hook("agent_prompt_prefix"),
to:
"prefix": "",

Because the new feature need to have access to working memory that it not already initialized when default prompt settings are declared. This change has no impact on normal operations of the cat.

UPDATED with new hook agent instruction definition

## Type of change

# Checklist:

### DIFF
--- a/core/cat/looking_glass/agent_manager.py
+++ b/core/cat/looking_glass/agent_manager.py
@@ -4,7 +4,7 @@ from copy import copy
 from cat.log import log
 from langchain.agents import AgentExecutor, ConversationalAgent
 from langchain.chains import LLMChain
-
+import re
 
 class AgentManager:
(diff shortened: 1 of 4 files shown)


pieroit:
Thanks!
In the future we can design the cat so prompt (and other settings) are orchestrated from db, websocket, endpoints and plugin in a consistent way. Too early for that, but we are on the right track!!

PR #332, opened 2023-07-03
Catch any exception when instantiating LLM

### LINKED ISSUE #323: [BUG] Setting wrong HuggingFace Hub models prevent the Cat from starting
**Describe the bug**
The HuggingFace Hub setting only accepts few categories of NLP models.
Setting a LLM from a not valid category saves the setting in the db, but then prevents the Cat from starting. Hence, changing back to another valid model is not possible anymore. The only solution is to delete `metadata-v3.db`.

**To Reproduce**
Steps to reproduce the behavior (as example):
1. Go to the `Setting` page
2. Configure LLM with `HuggingFace Hub`
3. Set the repo id from any model from HuggingFace with a tag different from  'text2text-generation', 'text-generation' or 'summarization' (e.g. use a 'Conversational' model like this: microsoft/DialoGPT-medium)
4. See error:
```bash
cheshire_cat_core           |   Got invalid task conversational, currently only ('text2text-generation', 'text-generation', 'summarization') are supported (type=value_error)
```
5. Change back to any other valid LLM (e.g. ChatGPT)
6. See error:
```bash
pydantic.error_wrappers.ValidationError: 1 validation error for HuggingFaceHub
cheshire_cat_core           | __root__
cheshire_cat_core           |   Got invalid task conversational, currently only ('text2text-generation', 'text-generation', 'summarization') are supported (type=value_error)
cheshire_cat_core           | 
cheshire_cat_core           | ERROR:    Application startup failed. Exiting.
```

**Expected behavior**
The Cat should check if the model is correct before getting the configuration and prevent setting it to no crash.
Additionally, the admin could notify the event.

**Additional context**
Specifically, the app crashes [here](https://github.com/cheshire-cat-ai/core/blob/1589af0638f9b584c9f568eba295e26c198b0efc/core/cat/mad_hatter/core_plugin/hooks/models.py#L50)

SantoSimone:
# Description

Fixes #323 
Includes simple logging as well

### DIFF
--- a/core/cat/mad_hatter/core_plugin/hooks/models.py
+++ b/core/cat/mad_hatter/core_plugin/hooks/models.py
@@ -53,7 +53,13 @@ def get_language_model(cat) -> BaseLLM:
         selected_llm_config = crud.get_setting_by_name(
             next(cat.db()), name=selected_llm_class
         )
-        llm = FactoryClass.get_llm_from_config(selected_llm_config.value)
+        try:
+            llm = FactoryClass.get_llm_from_config(selected_llm_config.value)
+        except Exception as e:
+            import traceback
+            traceback.print_exc()
+            llm = llms.LLMDefaultConfig.get_llm_from_config({})
+            
 
     return llm
 


pieroit:
thanks!

Issue #287, opened 2023-06-10
POST `rabbithole/memory` endpoint to upload exported memories


pieroit:
**Is your feature request related to a problem? Please describe.**
It is possible to export memories from the admin, and should be possible to upload them so people can share embeddings

**Describe the solution you'd like**
POST `rabbithole/memory` endpoint to upload exported JSON from memory

giovannialbero1992:
Hi @pieroit: this issue can be closed or is it missing something? Thanks

pieroit:
@giovannialbero1992 the endpoint is there but does nothing, it should be implemented

If you are interested in doing it I can give you some reference :)

giovannialbero1992:
Yes, for sure :)
I'm open to receive any reference.

pieroit:
I realized it is harder than I thought, because when you upload a document or an URL, the cat deals with text and in both cases he will store documents [via this function](https://github.com/cheshire-cat-ai/core/blob/main/core/cat/rabbit_hole.py#L230) that passes docs to memory, ending up [here](https://github.com/cheshire-cat-ai/core/blob/main/core/cat/memory/vector_memory.py#L76).

Note the `VectorMemoryCollection` class does not have the method directly, it is inherited from a [langchain class](https://github.com/hwchase17/langchain/blob/master/langchain/vectorstores/qdrant.py#L112) that deals with Qdrant.

In the case of memory upload, embeddings are already there and so we should [call directly Qdrant]() and pass metadata along. It is one of those cases were langchain is not enough :rabbit: 

It doesn't matter if you reach the end goal, a stub is enough so I can work on it or we do peer programming in open mic in the discord ;)

Let me know or we find something else to contribute, thanks again

giovannialbero1992:
Hi @pieroit sorry for being late, I'm available next week to go deeply on this issue.
In the meanwhile I worked to another issue (I'll reach you in the specific issue)

PR #362, opened 2023-07-15
`/rabbithole/memory` endpoint

nicola-corbellini:
# Description

This PR implements the `/rabbithole/memory` endpoint to upload vector memories from a JSON file to the declarative memory.
 
I also added a very basic test to check the endpoint status_code but it should be improved to test different embeddings than OpenAI and, of course, other tests are needed.

Related to issue #287 

## Type of change

# Checklist:


core/cat/routes/upload.py
```diff
@@ -115,9 +119,81 @@ async def upload_memory(
...
+
+    # Check the embedder used for the uploaded memories is the same the Cat is using now
+    upload_embedder = memories["embedder"]
+    cat_embedder = str(ccat.embedder.__class__.__name__)
+
+    if upload_embedder != cat_embedder:
+        raise HTTPException(
+            status_code=422,
+            detail={
+                "error": f'Embedder mismatch: uploaded file embedder {upload_embedder} is != from {ccat.embedder}'}
```
zAlweNy26, line 156: "is different" is better than "is !="

nicola-corbellini: Joking, I agree corrected

core/cat/routes/upload.py
```diff
@@ -115,9 +119,81 @@ async def upload_memory(
...
+    # check if MIME type of uploaded file is supported
+    if content_type != "application/json":
+        raise HTTPException(
+            status_code=422,
+            detail={
+                "error": f'MIME type {file.content_type} not supported. Admitted types: application/json'}
+        )
+
+    # Create temporary file
+    temp_file = tempfile.NamedTemporaryFile(dir="/tmp/", delete=False)
```
giovannialbero1992, line 134: I started to see these lines of code in several places, probably could be good start to move in a proper service these action in order to call directly a specific method and unify the logic in a one place.

giovannialbero1992: @nicola-corbellini what do you think about this :smiley:  ?

nicola-corbellini: Yes, I agree. It may be a good solution.
Yet, if the tmp file is truly needed here is still an open question to me honestly 😅

pieroit: @nicola-corbellini @giovannialbero1992 honestly I don't remember why we save the files on disk, it may be in fact unnecessary

nicola-corbellini: Alright @pieroit , I'll remove it then in the next commit. 
Also, as @giovannialbero1992 suggested, I could make a private method in the RabbitHole for mime type checking along with the `upload_memory` method we discussed.

pieroit: Sounds great!

core/cat/rabbit_hole.py
```diff
@@ -1,33 +1,114 @@
...
+
     def __init__(self, cat):
         self.cat = cat
 
+        self.file_handlers = {
+            "application/pdf": PDFMinerParser(),
+            "text/plain": TextParser(),
+            "text/markdown": TextParser(),
+            "text/html": BS4HTMLParser()
+        }
```
nicola-corbellini, line 35: sorry but here I don't remember how we got along. Let me know if you want this different

core/cat/rabbit_hole.py
```diff
@@ -1,33 +1,114 @@
...
+        upload_embedder = memories["embedder"]
+        cat_embedder = str(self.cat.embedder.__class__.__name__)
+
+        if upload_embedder != cat_embedder:
+            message = f'Embedder mismatch: file embedder {upload_embedder} is different from {self.cat.embedder}'
+            raise HTTPException(
+                status_code=422,
+                detail={
+                    "error": message
+                })
```
nicola-corbellini, line 70: here I had to raise an `HTTPException` from inside the `RabbitHole`. How could I keep the exception in the endpoint?

core/cat/rabbit_hole.py
```diff
@@ -1,33 +1,114 @@
...
+        # Check embedding size is correct
+        embedder_size = self.cat.memory.vectors.embedder_size
+        len_mismatch = [len(v) == embedder_size for v in vectors]
+
+        if not any(len_mismatch):
+            raise HTTPException(
+                status_code=422,
+                detail={
+                    "error": f'Embedding size mismatch: vectors length should be {embedder_size}'}
+            )
```
nicola-corbellini, line 94: same as above

core/cat/routes/upload.py
```diff
@@ -13,33 +13,35 @@
...
 
     # check if MIME type of uploaded file is supported
-    if content_type not in admitted_mime_types:
+    if content_type not in admitted_types:
         raise HTTPException(
-            status_code = 422,
-            detail = { "error": f'MIME type {file.content_type} not supported. Admitted types: {" - ".join(admitted_mime_types)}' }
+            status_code=422,
+            detail={
+                "error": f'MIME type {content_type} not supported. Admitted types: {" - ".join(admitted_types)}'}
```
nicola-corbellini, line 44: I don't know why GitHub says this line is conflicting

core/tests/routes/test_routes_upload.py
```diff
@@ -0,0 +1,90 @@
...
+        fake_memory = json.load(f)
+
+    response = client.post("/rabbithole/memory/",
+                           files={
+                               "file": ("sample.json",
+                                        fake_memory)
+                           },
+                           headers={"accept": "application/json"})
+
+    assert response.status_code == 200
```
nicola-corbellini, line 32: I think this test fails because an embedder or LLM should be set prior to uploading the memories, because of the checks about having the same embeder and embedding size

core/tests/routes/test_routes_upload.py
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    with pytest.raises(HTTPException) as e:
+        # Get response
+        _ = client.post("/rabbithole/memory/",
+                        files={
+                            "file": ("test_file.json",
+                                     json.dumps(fake_memory))
+                        },
+                        headers={"accept": "application/json"})
```
nicola-corbellini, line 90: I think both these tests fail because of the background task. Pytest tells the exceptions is correctly raised and detected, but complains that the response is already started, because is asynchronous

core/cat/rabbit_hole.py
nicola-corbellini, line 94: same as above

core/cat/rabbit_hole.py
nicola-corbellini, line 70: here I had to raise an `HTTPException` from inside the `RabbitHole`. How could I keep the exception in the endpoint?

core/cat/rabbit_hole.py
nicola-corbellini, line 35: sorry but here I don't remember how we got along. Let me know if you want this different

pieroit:
Thanks for your attention to details!!

Issue #366, opened 2023-07-18
DRY code


nicola-corbellini:
Following [this](https://github.com/cheshire-cat-ai/core/pull/362#discussion_r1267029114) suggestion from @giovannialbero1992,  I propose to aggregate memory specific [recalls from embedding](https://github.com/cheshire-cat-ai/core/blob/2f5c027dc80e0e1fdf5ec05e624a8c366619ebf9/core/cat/looking_glass/cheshire_cat.py#L199-L234) in a `for` loop for better organization and readability.

For example, we could have three lists with k, threshold and memory settings that could be zipped and looped.
E.g.:
```python
ks = [k_episodic, k_declarative, k_procedural]
thresholds = [t_episodic, ..., ...]
memories_settings = ["use_episodic_memory", ..., ...]

for k, threshold, setting in zip(ks, thresholds, memories_settings):

# use self.memory.vectors.episodic.recall_memories_from_embedding

``` 

Don't know if this could be useful or could be organized even better, I'm open to suggestions

Lv-291:
How about something like this?

```python
        memory_types = ["episodic", "declarative", "procedural"]

        for k, threshold, memory_type in zip(ks, thresholds, memory_types):
            setting = f"use_{memory_type}_memory"
            memory_key = f"{memory_type}_memories"

            if prompt_settings[setting]:
                # recall relevant memories
                vector_memory = getattr(self.memory.vectors, memory_type)

                memories = vector_memory.recall_memories_from_embedding(
                    embedding=memory_query_embedding,
                    k=k,
                    threshold=threshold,
                    metadata={
                        "source": user_id
                    }
                )
            else:
                memories = []

            self.working_memory[memory_key] = memories
```

nicola-corbellini:
@Lv-291 Exactly something like this. 

If `memories = []` is not needed further in the code I guess you could skip the `else` or even directly assign `self.working_memory[memory_key] = vector_memory.recall_memories_from_embedding`.

Btw let's hear also from @pieroit

pieroit:
Thanks @nicola-corbellini @Lv-291, I agree a refactor can be useful there.

Two observations:
- `metadata={ "source": user_id }` is only used for episodic memory, so there will be for each collection:
   - `embedding` (this should not change, but could if plugin devs decide to land a different query to each collection)
   - `k`: how many memories to recall
   - `threshold` minimal score
   - `metadata` conditions on recall 
[...]

PR #368, opened 2023-07-19
different hook for each memory

Lv-291:
# Description

- I removed the `before_cat_recalls_memories` hook.
- I created 3 hooks for each memory type
    - each hook returns a dictionary with:
      - `embedding`
      - `k`
      - `threshold`
      - `metadata`
     
Notes: the documentation of each hook needs to be updated, then I was thinking that since each hook has to return a dictionary with the same "format" maybe we could make a function that returns it 

- in `chesire_cat.py`  I do the execution of the 3 hooks (one for each memory)
- then in the `for` I check if the `embedding` parameter is empty, if it is it puts `memory_query_embedding`
- as arguments to the function `recall_memories_from_embedding()` I put the entire dictionary returned by the hooks

Related to issue #366 

## Type of change

# Checklist:

### DIFF
--- a/core/cat/looking_glass/cheshire_cat.py
+++ b/core/cat/looking_glass/cheshire_cat.py
@@ -183,9 +183,13 @@ class CheshireCat:
         user_message = self.working_memory["user_message_json"]["text"]
         prompt_settings = self.working_memory["user_message_json"]["prompt_settings"]
 
-        # hook to do something before recall begins
-        k_episodic, threshold_episodic, k_declarative, threshold_declarative, k_procedural, threshold_procedural = self.mad_hatter.execute_hook(
-            "before_cat_recalls_memories", user_message)
+        # hooks to do something before recall begins
+        memory_parameters = [
+            self.mad_hatter.execute_hook("before_cat_recalls_episodic_memories", user_message, user_id),
+            self.mad_hatter.execute_hook("before_cat_recalls_declarative_memories", user_message),
+            self.mad_hatter.execute_hook("before_cat_recalls_procedural_memories", user_message)]
+
+        memory_types = ('episodic', 'declarative', 'procedural')
 
         # We may want to search in memory
         memory_query_text = self.mad_hatter.execute_hook("cat_recall_query", user_message)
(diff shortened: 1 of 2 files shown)


pieroit:
Thanks @Lv-291 

Ok for the `for` loop.

I propose to have the hooks take a default dictionary as default, so hooks in the future can be piped (different hooks can examine and modify the dictionary in a sequence)

The hooks are called like this from the cat:

```python
default_episodic_recall_config = {
    "embedding": [0.2, 3.1, 3.1],
    "k": 3,
    "threshold": 0.7,
    "metadata": None,
}

episodic_recall_config = self.mad_hatter.execute_hook(
    "before_cat_recalls_episodic_memories", default_episodic_recall_config
)
```


Also, let's leave the `before_cat_recalls_memories` hook, no inputs or outputs, in case somebody needs to do something in a plugin before the recall process starts:
```python
self.mad_hatter.execute_hook("before_cat_recalls_memories")
```

Tell me if you need clarifications! Don't worry if the PR is not perfect, I'll fix what needs to be fixed ;)

@nicola-corbellini @EugenioPetulla your take on this, it should be in line with the proposals we had recently :D

Lv-291:
I didn't quite understand the "pipeline" thing. Do you mean to do something like this?
```python
default_episodic_recall_config = {
    "embedding": [0.2, 3.1, 3.1],
    "k": 3,
    "threshold": 0.7,
    "metadata": None,
}

episodic_recall_config = self.mad_hatter.execute_hook(
    "before_cat_recalls_episodic_memories", default_episodic_recall_config
)

declarative_recall_config = self.mad_hatter.execute_hook(
    "before_cat_recalls_declarative_memories", episodic_recall_config
)
```
If so, for example in the episodic memory hook there should be something like this:
```python
default_episodic_recall_config["metadata"] = {"source": user_id}
return default_episodic_recall_config
```
Did I understand correctly?

pieroit:
@Lv-291 each collection has its own default, the pipeline is for different plugins that attach to the same hook. Those hooks take the dict as input and return it as output.

So there is a `default_episodic_recall_config`, a `default_declarative_recall_config` and a `default_procedural_recall_config`.

What do you think? Is it clear?

Lv-291:
@pieroit I get it now. So the hooks for now should just return the dictionary passed as input.
Do I put all the "`default_recall_configs`" inside the `recall_relevant_memories_to_working_memory()` method or is there something that handles the default configurations?

pieroit:
> @pieroit I get it now. So the hooks for now should just return the dictionary passed as input.

Yes kudos ;)

> Do I put all the "`default_recall_configs`" inside the `recall_relevant_memories_to_working_memory()` method or is there something that handles the default configurations?

Just put them in the method thanks

Lv-291:
okay thank you. Just one last question, do the memory hooks still need`user_message` as an argument? Previously `before_cat_recalls_memories` had it but did not use it.

pieroit:
> okay thank you. Just one last question, do the memory hooks still need`user_message` as an argument? Previously `before_cat_recalls_memories` had it but did not use it.

Not necessary, because user_message can also be found in working memory

Lv-291:
I made the changes, feel free to edit the documentation I put, the one in `recall_relevant_memories_to_working_memory()` needs to be changed I think.

pieroit:
Nice! Thanks ;)

PR #386, opened 2023-07-31
Refine tools output

Pingdred:
# Description

The tools output has been added in the memory chain prompt so it can be enriched with memory and the cat role.

## Type of change

# Checklist:


cristianorevil:
@Pingdred i suggest to make the phrase "## Tools output: \n" overridable by hooks, so can be translated or maybe modified, what do you think about?

pieroit:
> @Pingdred i suggest to make the phrase "## Tools output: \n" overridable by hooks, so can be translated or maybe modified, what do you think about?

let's make the agent better and after it stabilizes we can make all prompts (and prompt parts) customizable ;)

Issue #384, opened 2023-07-29
Default embedder should at least be a one hot encoder


pieroit:
At the moment the default embedder just throws numbers at random :@

It could be a good idea to have as default embedder a OneHotEncoder from scikit-learn (which is already included as a core dependency) wrapped in a langchain.Embeddings subclass.

We can use as vocabulary the list of tokens used by openAI, or any widely used token list, so the embedder is not dependent on any language.

nickprock:
I can working on

nickprock:
Hi @pieroit @nicola-corbellini this is a function for very sparse embedder

```
import tiktoken
def defaultSparseEmbedder(text):
    enc = tiktoken.get_encoding("p50k_base")
    idx_list = enc.encode(text)
    emb = [0]*enc.n_vocab
    for idx in idx_list:
        emb[idx]+=1
    return emb
```

PR #397, opened 2023-08-04
Dumb embedder

nicola-corbellini:
# Description

> "The bumblebee's wing structure, in relation to its weight, is not suitable for flight, but it does not know this and flies anyway."
\- *Albert Einstein*

Similarly, this embedder is not a model properly trained and this makes it not suitable to effectively embed text, "*but it does not know this and embeds anyway*".

Related to issue #384 

P.S.: the quote is **fake**

## Type of change

# Checklist:


core/cat/factory/embedder.py
```diff
@@ -83,6 +96,7 @@ class Config:
 
 
 SUPPORTED_EMDEDDING_MODELS = [
+    EmbedderDumbConfig,
     EmbedderFakeConfig,
```
nicola-corbellini, line 100: Should I make the fake embedder invisible here?

pieroit:
Amazing PR

Is it worth it to [normalize the embeddings](https://stackoverflow.com/a/56284637) so they are all of lenght 1?

BTW there is something wrong with the projection in memory view, or we miss something, because the red dot is always distant from the others (even with embeddings of comparable norm)

... embeds anyway

PR #404, opened 2023-08-07
Really fixed the return direct

Pingdred:
# Description

The return direct should be really fixed now, when a tool having `return_direct = True` is used the `execute_tool_agent` return immediately and the tool output is showed in chat.

If multiple tools are used, the `intermediate_steps` are still shown in the `why` but in the chat only the output of the tool with `return_direct` is shown.

## Type of change

# Checklist:


pieroit:
Thanks @Pingdred !

@cristianorevil issue about `return_direct` in develop is fixed, let us know ;)

PR #406, opened 2023-08-08
Create tag workflow

nicola-corbellini:
# Description

Workflow to create a tag on pushes on the main branch.

## Type of change

# Checklist:

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -1,0 +1,37 @@
+name: Cheshire Cat Action create tag
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


pieroit:
So if I get it right, this always updates the latest release and creates a new one when version in the toml changes?
Keeping an archive of the old ones?

nicola-corbellini:
Roughly yes, actually this doesn't create a release but only a tag. If the tag already exists, it pushes on the same, otherwise it creates a new one. The tag must be updated manually in the toml. 

We used only the tag 'cause we don't really have something to release. The tag automatically creates the .zip and .tar.gz. As far as I've understood, in the release you should also put the binaries of the app.

PR #418, opened 2023-08-18
Fixed `LLMCustom` setup

nicola-corbellini:
# Description

This PR fixes a couple of issues related to using a custom LLM.

## Type of change

# Checklist:


core/cat/factory/llm.py
```diff
@@ -46,7 +46,7 @@ class LLMCustomConfig(LLMSettings):
     @classmethod
     def get_llm_from_config(cls, config):
         # options are inserted as a string in the admin
-        if type(config["options"]) == str:
+        if isinstance(config["options"], dict):
```
pieroit, line 49: Here we were checking for a string because there is a free text input box in the admin (where people can write a doctionary). That dictionary is later parsed with `json.loads`

nicola-corbellini: Ah I've misunderstood that, I'm correcting it.

Also, I think I've slipped and branched out main rather than develop. It's better if I open a new PR, isn't it?

pieroit: They should be aligned apart from the readme, don't worry in this case nothing happened

PR #441, opened 2023-08-29
Refactored WS messages functionality

zAlweNy26:
# Description

I think I improved the way to send ws messages, so that if in the future it should change, we only have to change that specific function. In this way also it's easier for plugins to send directly messages to the chat without waiting for a user input that triggers the tool each time.

## Type of change

# Checklist:

### DIFF
--- a/core/cat/rabbit_hole.py
+++ b/core/cat/rabbit_hole.py
@@ -217,35 +217,18 @@ class RabbitHole:
         blob = Blob(data=file_bytes,
                     mimetype=content_type,
                     source=source).from_data(data=file_bytes,
-                                             mime_type=content_type)
+                                             mime_type=content_type,
+                                             path=source)
         # Parser based on the mime type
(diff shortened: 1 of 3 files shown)


core/cat/rabbit_hole.py
```diff
@@ -217,36 +217,19 @@ def file_to_docs(
         blob = Blob(data=file_bytes,
                     mimetype=content_type,
                     source=source).from_data(data=file_bytes,
-                                             mime_type=content_type)
+                                             mime_type=content_type,
+                                             path=source)
```
zAlweNy26, line 221: Please don't kill me Piero, I also added the path here because it doesn't arrive any source to the parsers. As the [documentation](https://api.python.langchain.com/en/latest/document_loaders/langchain.document_loaders.blob_loaders.schema.Blob.html#langchain.document_loaders.blob_loaders.schema.Blob.from_data) says, `path` will set the source of the Blob

PR #446, opened 2023-09-01
Updated hook decorator

zAlweNy26:
# Description

In this way, a plugin developer can either define the hook using the function name or using the `name` argument of the @hook with the help of the autocomplete thanks to the Literal `CAT_HOOKS`.

Thanks to this PR, we still support the old way of defining hooks, but also adding a new and easier way to define them.

## Type of change

# Checklist:

### DIFF
--- a/core/cat/looking_glass/agent_manager.py
+++ b/core/cat/looking_glass/agent_manager.py
@@ -1,7 +1,7 @@
 from langchain.prompts import PromptTemplate
 from langchain.chains import LLMChain
 from langchain.agents import AgentExecutor, LLMSingleActionAgent
-
+import traceback
 from cat.looking_glass.prompts import ToolPromptTemplate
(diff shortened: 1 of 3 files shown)


zAlweNy26:
@pieroit @nicola-corbellini what do you think about this?

pieroit:
> @pieroit @nicola-corbellini what do you think about this?

I'll leave the PR open, but I'm not convinced it is worth it to have this just for autocomplete

zAlweNy26:
But nothing breaks, everything continues to work as before, there is simply this new extra mode for those who want to 😁

nicola-corbellini:
apart from merging this or not, I'd populate the `Literal` list programmatically rather than hard coding the hooks' names

zAlweNy26:
How can it be done?

PR #448, opened 2023-09-01
Fixed workflows

zAlweNy26:
# Description

In this way, when a workflow of the same group has to start, it stops the one that is running before, decreasing the load on the poor Github (sorry Microsoft).

For example, if a PR adds 2 new commits one after another in just a few seconds, the one running before will be stopped. It is useless to see the results of that one as the new one is more important.

## Type of change

# Checklist:


nicola-corbellini:
Thank you 😅

PR #457, opened 2023-09-18
Add list of hooks and tools when fetching plugins

zAlweNy26:
# Description

In this way, you can visualize in the frontend or from a generic client, the used hooks by each plugin and their priority, for a better understanding of the flow of the plugins you installed. You can also see the tools created by the plugins.

## Type of change

# Checklist:

### DIFF
--- a/core/cat/routes/plugins.py
+++ b/core/cat/routes/plugins.py
@@ -40,8 +40,8 @@ async def get_available_plugins(
         # get manifest
         manifest = deepcopy(p.manifest) # we make a copy to avoid modifying the plugin obj
         manifest["active"] = p.id in active_plugins # pass along if plugin is active or not
-        manifest["used_hooks"] = [{ "name": hook.name, "priority": hook.priority } for hook in p.hooks]
-        manifest["used_tools"] = [tool.func.__name__ for tool in p.tools]
+        manifest["hooks"] = [{ "name": hook.name, "priority": hook.priority } for hook in p.hooks]
+        manifest["tools"] = [{ "name": tool.name } for tool in p.tools]
         
         # filter by query
         plugin_text = [str(field) for field in manifest.values()]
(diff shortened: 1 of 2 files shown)


pieroit:
This is a good idea thanks
What about adding the tools?

zAlweNy26:
I thought about that too, but... who cares about the tools names? The description of the plugin should already answer a hypothetical question like "how do I trigger this plugin?", or am I wrong?

core/cat/routes/plugins.py
```diff
@@ -40,6 +40,8 @@ async def get_available_plugins(
         # get manifest
         manifest = deepcopy(p.manifest) # we make a copy to avoid modifying the plugin obj
         manifest["active"] = p.id in active_plugins # pass along if plugin is active or not
+        manifest["used_hooks"] = [{ "name": hook.name, "priority": hook.priority } for hook in p.hooks]
+        manifest["used_tools"] = [tool.func.__name__ for tool in p.tools]
```
zAlweNy26, line 44: Also added tools, should the entry be called `used_tools`, `tools` or `created_tools`?

zAlweNy26:
I was thinking about changing the data structure in this way, where the key is the priority:

```json
{
  "0": ["hook name 1", "hook name 2"],
  "1": ["hook name 3", "hook name 4"]
}
```

(Suggested by the Cat 😉)

pieroit:
> I was thinking about changing the data structure in this way, where the key is the priority:
> 
> ```json
> {
>   "0": ["hook name 1", "hook name 2"],
>   "1": ["hook name 3", "hook name 4"]
> }
> ```
> 
> (Suggested by the Cat 😉)

The previous was better, because it makes no assumptions on ordering or indexing.
Two keys `tools` and `hooks` with a list of dicta inside

Issue #461, opened 2023-09-18
[BUG] application crashes if an OpenAI API key with exceeded quota is provided in embeddings settings


zenimarc:
**Describe the bug**
If you exceed the OpenAI API quota, (in my case I generated an api key but no payment method is registered), and you put this api key in the embeddings configuration. when you save, a loading spinner is shown but it gets stuck, from the console I can see this error printed multiple times:
```
Retrying langchain.embeddings.openai.embed_with_retry.<locals>._embed_with_retry in 4.0 seconds as it raised RateLimitError: You exceeded your current quota, please check your plan and billing details..
cheshire_cat_core           | Retrying langchain.embeddings.openai.embed_with_retry.<locals>._embed_with_retry in 4.0 seconds as it raised RateLimitError: You exceeded your current quota, please check your plan and billing details..
cheshire_cat_core           | Retrying langchain.embeddings.openai.embed_with_retry.<locals>._embed_with_retry in 4.0 seconds as it raised RateLimitError: You exceeded your current quota, please check your plan and billing details..
```
Even if i restart the project with docker during startup it continues to say the same previous error and the following error is shown:  ERROR:    Application startup failed. Exiting.

**To Reproduce**
1. fresh installation
2. go to settings
3. embeddings settings
4. openAI
5. put an api key which has no quota left
6. save

**Expected behavior**
print you exceeded your quota and let you input another key or change embeddings

**Additional context**

pieroit:
Thanks, we know very well this error but never managed it.
Are you able to do a PR?

kodaline:
@pieroit in case I can think on a general solution for this one

pieroit:
@kodaline thanks! If you look around in the code you will find consistency in how exceptions are thrown back to the client

Let me know if you need clarification or reference

PR #470, opened 2023-09-27
Fixed the rate limit issue for openai embedder

mmassime:
# Description

Previously there was no check to make sure that the openai key used when setting up an openai embedder was not expired. I added two checks:

1. When inserting the key: an error will be shown and the user will have to insert a new key.
2. Every time the cat is instantiated: this is done becuase the user can still insert a valid key which could expire in the following days and this could cause problems when the cat is launched. To avoid this, the check is re-executed at any instantiation of the cat and if a non-valid key is detected then the embedder is set to DumbEmbedder and a warning is logged in the console.

Related to issue #461 

## Type of change

# Checklist:


pieroit:
Closing this to work on a more general, non openai-specific solution.
Thanks for this anyway :D

PR #478, opened 2023-10-03
OpenAI temperature setting added

avvrossini:
# Description

I've implemented the feature that allows users to set the "temperature" in requests made via OpenAI's API. I've set a default value of 0.7.

Temperature is a parameter that influences the creativity of the model's responses:

Lower values (e.g., 0.2) make the output more deterministic and closely tied to the context of the provided prompt.
Higher values (e.g., 0.8 or above) make the output more random and creative, potentially deviating from the prompt's context.
While in the past the acceptable range for temperature was between 0.0 and 1.0, it now seems that OpenAI allows values up to 2.0. However, there's some ambiguity in OpenAI's documentation as it mentions that values above 0.8 would produce more random outputs, but the default is now set to 1.

Given this uncertainty, I've chosen not to impose a maximum limit on the temperature parameter. I suggest monitoring any updates from OpenAI's documentation and, if necessary, adjust our default value or the accepted range at a later date.

## Type of change

# Checklist:

### DIFF
--- a/core/cat/factory/llm.py
+++ b/core/cat/factory/llm.py
@@ -84,6 +84,7 @@ class LLMLlamaCppConfig(LLMSettings):
 class LLMOpenAIChatConfig(LLMSettings):
     openai_api_key: str
     model_name: str = "gpt-3.5-turbo"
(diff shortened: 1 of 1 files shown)


pieroit:
Thanks! And welcome :)

Issue #488, opened 2023-10-17
[BUG] Websocket messages are mixed


Pingdred:
**Describe the bug**
At the moment there is a unique [queue](https://github.com/cheshire-cat-ai/core/blob/1e72089959efc5342d06d6a4e1c840c811efe469/core/cat/looking_glass/cheshire_cat.py#L74C29-L74C29) for all the ws messages, so if the are multiple websocket connection open is not predictable to which connection the messages are sent, because each ws connection read from the same queue.

**To Reproduce**
1. Open two ws connection, ad example open the admin in two different browser tab;
2. Do something that send a notification, like upload a document;

**Expected behavior**
Notifications sent while the document uploading should only be shown in the tab where the document is uploaded.

A possible solution is to make the queue](https://github.com/cheshire-cat-ai/core/blob/1e72089959efc5342d06d6a4e1c840c811efe469/core/cat/looking_glass/cheshire_cat.py#L74C29-L74C29) a dictionary of queue using as key for each queue the user id.

pieroit:
@AlboCode please comment here so I can assign also to you :)

AlboCode:
here I am 😄

AlboCode:
I have some doubt if the user_id is the solution. When a document is uploaded, we don't know where is the connected user 🤔 
Only when a message is sent on the WS there is also the user_id property.
Is there a way to set a "SESSION_ID" in the header when the user connect to the websocket?

pieroit:
> I have some doubt if the user_id is the solution. When a document is uploaded, we don't know where is the connected user 🤔 Only when a message is sent on the WS there is also the user_id property.

At the moment declarative memory is general (not segregated by users) but we can easily extend the http endpoints to accept a `user_id` param. (As it happens in the `memory/recall` endpoint)
This tests make it clear:
https://github.com/cheshire-cat-ai/core/blob/09d5e5c6e5b91e919c740fc7ed434854c35f7bde/core/tests/routes/memory/test_memory_by_user.py

> Is there a way to set a "SESSION_ID" in the header when the user connect to the websocket?
I don't understand this can you clarify?

AlboCode:
my doubt was, when there is just a connected web socket listening for notifications how do we know the connected user_id? we send all the notifications to the default user id (user) only in a broadcast way? 
maybe I'm missing some parts

giovannialbero1992:
@AlboCode the message is already sent in a proper way with `send_personal_message` method.
[...]

PR #509, opened 2023-10-20
Added user_id to websocket route

zAlweNy26:
# Description

With this PR, the WebSocket messages should be split for each user_id. When the user_id is not passed, it will default to "user".

This actually breaks token streaming.

Related to issue #488 

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

### DIFF
--- a/core/cat/looking_glass/cheshire_cat.py
+++ b/core/cat/looking_glass/cheshire_cat.py
@@ -1,7 +1,8 @@
 import time
 from copy import deepcopy
 import traceback
-from typing import Literal, get_args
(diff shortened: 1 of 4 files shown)


zAlweNy26:
@Pingdred @nicola-corbellini when two websockets connects, everything it's ok. But when one of the two users try to send a message, I get this error:
`RuntimeError: cannot call recv while another coroutine is already waiting for the next message`

zAlweNy26 (commented):
fixed in the last commit

Issue #514, opened 2023-10-22
[BUG] Plugin files' names can't start with ".py" 


nicola-corbellini:
**Describe the bug**
The Cat is not able to load plugins that have modules whose names start with ".py" because, when importing those modules, the Cat replaces ".py" with an empty string (see [here](https://github.com/cheshire-cat-ai/core/blob/464bbe3cbc9a6c544701f1ec478a120df3f288d7/core/cat/mad_hatter/plugin.py#L37)).

E.g. a plugin `cat/plugins/pythonic/pythonic.py` breaks the plugin activation because the imported module becomes `cat.plugins.pythonicthonic.

**To Reproduce**
Steps to reproduce the behavior (as exampple`):
1. Create a plugin file named `pythonic/pythonic.py`
2. Enable it

pieroit:
Should suffice to substitute [here](https://github.com/cheshire-cat-ai/core/blob/0334330475254817c8480644c385975f37a42409/core/cat/mad_hatter/plugin.py#L230C13-L230C71):
`py_filename = py_file.replace("/", ".").replace(".py", "")`
with:
`py_filename = py_file.replace(".py", "").replace("/", ".")`

PR #518, opened 2023-10-23
Fix dynamic plugin import

zAlweNy26:
# Description

With this PR I fixed the import for the plugins that was causing a bug.

Related to issue #514 

## Type of change

# Checklist:

### DIFF
--- a/core/cat/mad_hatter/plugin.py
+++ b/core/cat/mad_hatter/plugin.py
@@ -3,7 +3,7 @@ import sys
 import json
 import glob
 import traceback
-import imp
+from importlib import machinery
 from typing import Dict
 from inspect import getmembers
 from pydantic import BaseModel
@@ -213,7 +213,7 @@ class Plugin:
 
             # save a reference to decorated functions
             try:
-                plugin_module = imp.load_source(module_name, py_file)
+                plugin_module = machinery.SourceFileLoader(module_name, py_file).load_module()
                 hooks += getmembers(plugin_module, self._is_cat_hook)
                 tools += getmembers(plugin_module, self._is_cat_tool)
                 plugin_overrides += getmembers(plugin_module, self._is_cat_plugin_override)
--- a/core/cat/mad_hatter/plugin.py
+++ b/core/cat/mad_hatter/plugin.py
@@ -2,8 +2,8 @@ import os
 import sys
 import json
 import glob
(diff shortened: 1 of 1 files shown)


core/cat/mad_hatter/plugin.py
```diff
@@ -207,18 +207,18 @@ def _load_decorated_functions(self):
...
-            py_filename = py_file.replace("/", ".").replace(".py", "")  # this is UGLY I know. I'm sorry
+            module_name = os.path.splitext(os.path.basename(py_file))[0]
 
-            log.info(f"Import module {py_filename}")
+            log.info(f"Import module {py_file}")
 
             # save a reference to decorated functions
             try:
-                plugin_module = importlib.import_module(py_filename)
+                plugin_module = imp.load_source(module_name, py_file)
```
nicola-corbellini, line 216: why not importlib?

zAlweNy26: I didn't find a clean way to do it with importlib. If you have a better way to do that, I accept suggestions 👍🏽

nicola-corbellini: Can you try something like this, please?
```python
spec=importlib.util.spec_from_file_location(module_name, py_file)
plugin_module = importlib.util.module_from_spec(spec)
```
I think `imp` package is deprecated

zAlweNy26: I tried that way but it says it doesn't find any hooks

zAlweNy26: @nicola-corbellini I found a way to use importlib, if you want to check it out

core/cat/mad_hatter/plugin.py
```diff
@@ -213,7 +213,7 @@ def _load_decorated_functions(self):
 
             # save a reference to decorated functions
             try:
-                plugin_module = imp.load_source(module_name, py_file)
+                plugin_module = machinery.SourceFileLoader(module_name, py_file).load_module()
```
nicola-corbellini, line 216: Top level Py-dev

zAlweNy26: Are you trolling me?

nicola-corbellini: I'm greeting you

PR #525, opened 2023-10-23
FIX: docker build

kodaline:
# Description

## Issue

The pre-built docker image contains only `cat/plugins` folder instead all the core. This cause the current pre-built image by github action to be incomplete and to fail with error `/usr/local/bin/python: No module named cat.main`.

## Solution

Change COPY command of the Dockerfile to include the whole `cat` folder and modify the docker-compose.yml accordingly to use the generated image and to not bind mount `/core` anymore.

## Extra

Edit github action to support multi-platform build.

## Type of change

# Checklist:

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -36,6 +36,17 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
+
+      # Uses the `docker/setup-qemu-action@v3`
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      # Uses the `docker/setup-buildx-action@v3`
+      - name: Set up docker Buildx
+        uses: docker/setup-buildx-action@v3
+        with:
+          platforms: linux/amd64,linux/arm64
+
       # Uses the `docker/login-action`
       # action to log in to the Container registry using the account and password that will publish the packages.
       # Once published, the packages are scoped to the account defined here.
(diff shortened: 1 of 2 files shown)


kodaline:
@nicola-corbellini Hi, here is the pull-request, not sure about the build action, I never tested one.
Please give it a look when you can

pieroit:
Hi @kodaline, welcome and thanks :)
I'll leave a review

docker-compose.yml
```diff
@@ -3,8 +3,7 @@ version: '3.7'
 services:
 
   cheshire-cat-core:
-    build:
-      context: ./core
+    image: ghcr.io/cheshire-cat-ai/core:latest
```
pieroit, line 6: The already built image is meant to be used outside of this repo:
- if you want to `develop` on cat core, you git pull the code and build the image locally (in this case the image opens a volume on the code which is downloaded and versioned via git)
- if you want to just `use` the cat, you directly run the ghcr.io docker image, opening a volume just on the plugin folder and eventually some other folders

I don't know if that is clear, otherwise I'll reword :)

docker-compose.yml
```diff
@@ -22,8 +21,6 @@ services:
       - SAVE_MEMORY_SNAPSHOTS=${SAVE_MEMORY_SNAPSHOTS:-false}
     ports:
       - ${CORE_PORT:-1865}:80
-    volumes:
-      - ./core:/app
```
pieroit, line 26: This volume is necessary here, as the code is obtained via git. Outside of this repo, it make more sense to use the built image (already containing the cat code) and opening a volume just on plugins

pieroit (changes requested):

docker-compose.yml
kodaline: As discussed privately, the file is ok as it is

docker-compose.yml
kodaline: As above

pieroit:
Thanks and welcome

PR #567, opened 2023-11-15
add working memory for the agent prompt prefix and suffix hook

AlboCode:
# Description

Add working memory to prompt flow suffix and prefix hooks 

Related to issue #(issue)

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:


pieroit:
@AlboCode after sessions are introduced you will find the correct working memory (user/session specific) in `cat.working_memory`

We need a little more time!

Also we cannot change hooks so easily, many plugins rely on that signature

PR #574, opened 2023-11-18
Speed up build time when only plugins requirements installation

sambarza:
# Description

- Removed mkdir of folders
The WORKDIR command already create the folder if missing

- Speed up build time when only plugins requirements installation

The dev define a new package in the requirements file.
The dev run the build with `docker compose build`.
The build recreate all the layers.
If the layer is the same, docker keep it, when a layer change, docker create the layer and all the layer after it.

Moved the apt-get commands before the others steps.
Moved the commands to copy of plugins and installation of plugins dependencies as last commands.

Related to issue, not related to issue

## Type of change

# Checklist:

### DIFF
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -1,11 +1,6 @@
 
 FROM python:3.10.11-slim-bullseye
 
-### PREPARE BUILD WITH NECESSARY FILES AND FOLDERS ###
-COPY ./pyproject.toml /app/pyproject.toml
-COPY ./cat /app/cat
-COPY ./install_plugin_dependencies.py /app/install_plugin_dependencies.py
-
 ### SYSTEM SETUP ###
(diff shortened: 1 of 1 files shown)


core/Dockerfile
```diff
@@ -1,27 +1,30 @@
...
+
+### COPY PLUGINS ###
+COPY ./cat/plugins /app/cat/plugins
+
+### INSTALL PYTHON DEPENDENCIES (Plugins) ###
+COPY ./install_plugin_dependencies.py /app/install_plugin_dependencies.py
+RUN python3 install_plugin_dependencies.py
 
 ### FINISH ###
-CMD python3 -m cat.main
```
pieroit, line 27: We need this CMD in case you launch the image solo (out of docker compose)

pieroit: Please the PR branched from `develop` and towards `develop`
(Because there are commits on `main` that are not on `develop` and vice versa)

PR #576, opened 2023-11-18
Speed up build plugins

sambarza:
# Description

- Removed mkdir of folders

The WORKDIR command already create the folder if missing

- Speed up build time when only plugins requirements installation

The dev define a new package in the requirements file.
The dev run the build with `docker compose build`.
The build recreate all the layers.
If the layer is the same, docker keep it, when a layer change, docker create the layer and all the layer after it.

Moved the apt-get commands before the others steps.
Moved the commands to copy of plugins and installation of plugins dependencies as last commands.
In this way, when a dev change the requirements file and do `docker compose build`, only the `COPY PLUGINS` and `INSTALL PYTHON DEPENDENCIES (Plugins)` are done, no time spent to do apt-get anymore.

Not related to issue

## Type of change

# Checklist:

### DIFF
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -1,11 +1,6 @@
 
 FROM python:3.10.11-slim-bullseye
 
-### PREPARE BUILD WITH NECESSARY FILES AND FOLDERS ###
-COPY ./pyproject.toml /app/pyproject.toml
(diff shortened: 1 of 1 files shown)


pieroit:
Thanks! Way faster

PR #588, opened 2023-11-24
IMP: method ad hoc for max user input

kodaline:
## Improvement for #558 

Added ad hoc method to manage very long input messages

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas

PR #612, opened 2023-12-07
imp: ad hoc method for max user input

kodaline:
## Description

Added ad hoc method to manage very long input messages

Improvement for issue #588 

## Type of change

# Checklist:

### DIFF
--- a/core/cat/looking_glass/stray_cat.py
+++ b/core/cat/looking_glass/stray_cat.py
@@ -233,18 +233,8 @@ class StrayCat:
 
             # TODO: move inside a function
             # split text after MAX_TEXT_INPUT tokens, on a whitespace, if any, and send it to declarative memory
-            if len(self.working_memory["user_message_json"]["text"]) > MAX_TEXT_INPUT:
-                index = MAX_TEXT_INPUT
-                char = self.working_memory["user_message_json"]["text"][index]
-                while not char.isspace() and index > 0:
-                    index -= 1
-                    char = self.working_memory["user_message_json"]["text"][index]
-                if index <= 0:
-                    index = MAX_TEXT_INPUT
-                self.working_memory["user_message_json"]["text"], to_declarative_memory = self.working_memory["user_message_json"]["text"][:index], self.working_memory["user_message_json"]["text"][index:]
-                docs = self.rabbit_hole.string_to_docs(stray=self, file_bytes=to_declarative_memory, content_type="text/plain")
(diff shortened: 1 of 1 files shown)


kodaline:
@pieroit do not merge this yet, I want to make some changes

pieroit:
> @pieroit do not merge this yet, I want to make some changes

Ok!

Pingdred:
I don't know if I understand correctly, but this way if the user's question is too long it will be sent down the rabbit hole but the agent will continue to work as usual, what is the reason for doing this?

kodaline:
> I don't know if I understand correctly, but this way if the user's question is too long it will be sent down the rabbit hole but the agent will continue to work as usual, what is the reason for doing this?

this was done to avoid crashes in case of sending too long messages, that makes openai crash for example, because the model accepts less tokens

pieroit:
@Pingdred @kodaline let's merge it, and later on design a way to jump directly to final response without involving nor the agent, nor recall. May be useful for a reflex-like response in a lot of situations, also for plugins

Pingdred:
Yes, I remember the problem, but from what I have seen the user message is not changed, so the agent input is still the same, am I missing something?

kodaline:
> @Pingdred @kodaline let's merge it, and later on design a way to jump directly to final response without involving nor the agent, nor recall. May be useful for a reflex-like response in a lot of situations, also for plugins

I can rework this one, as you prefer

kodaline:
> Yes, I remember the problem, but from what I have seen the user message is not changed, so the agent input is still the same, am I missing something?

I know, I noticed that too, but initially the idea was to do it as it is now, actually I wanted to talk to Piero about this XD

pieroit:
@kodaline yes it is perfect! Thanks

kodaline:
> @kodaline yes it is perfect! Thanks

ok, will rework this one directly, I think it's the best thing to do
would you briefly explain the new behaviour?

pieroit:
@kodaline sorry I'll explain myself better

Since the "reflex" must be designed and should offer a new hook, I prefer this PR as is and later on have a talk

pieroit:
Thanks @kodaline :dagger:
@Pingdred I got only now what you meant, fixed [here](https://github.com/cheshire-cat-ai/core/blob/000eeba3b8e456e838ab5f98f44be8a99d8824b3/core/cat/looking_glass/stray_cat.py#L333)

When you are free, we can design the mitical `reflex` hook

PR #634, opened 2023-12-25
add context and fix git command in first code contribution

NJacksonDev:
# Description

In the Contributing.md file, I added some additional context to #1 under "Your First Code Contribution" (which I think is helpful for people getting started with contributing) and also fixed the git command in #2 (instead of `git checkout -b develop`, to switch to the branch and not create a new develop branch, it should be `git checkout develop`)


pieroit:
Welcome @NJacksonDev, thanks :D

PR #740, opened 2024-03-09
Run Cheshire without docker

paoloach:
# Description

## Motivation
As developer,  when I develop, it seems much simpler to me running the cat  inside a virtual environment  on my host machine instead to run in a docker machine. In this way I can leverage on tools  on my IDE and I can debug without making strange hack. On the contrary, I see any benefit to develop using a  docker machine.

## Purpose of the change
I made some changes in the code that didn't allow it to run in a virtual enviroment without breacking the previous behavior  and add documentation on how to use it

## Implementations
Because the problem is a few fixed path in the codebase,  I change this values with the one get from environment variables in order to allow the developer to choice that working folder and the server port, using as default the previous ones.

Add also a brief description on how to use this features.

## Type of change

# Checklist:

### DIFF
--- a/core/.gitignore
+++ b/core/.gitignore
@@ -1,0 +1,2 @@
+/build
+/Cheshire_Cat.egg-info
--- a/core/cat/log.py
+++ b/core/cat/log.py
@@ -219,8 +219,9 @@ class CatLogEngine:
         if secure != '':
             secure = 's'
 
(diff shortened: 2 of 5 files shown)


pieroit:
@paoloach too soon for this, thanks for your PR but I cannot yet merge it.
It's going to be way more complicated than this (and also we need to get ready to support noobs on their python environment)

PR #742, opened 2024-03-09
Develop

MaxDam:
I have created two methods on a new file form_utils.py:
-classify (receives the user message and a descriptive enum, and based on the message chooses the desired entry);
-call_llm_and_extract_json (unique method that uses LLChain and other transformations to capture the json from the prompt).

I also updated CatForm so that it calls the call_llm_and_extract_json method explained above, to invoke llm and parse response in a right way ; and I have slightly modified the check_exit_prompt prompt, adding the phrase "If you are not sure answer `false`." which helps small models better understand what to do.

(In this mode catForm seems to work also with ollama gemma:2B)

### DIFF
--- a/core/cat/mad_hatter/decorators/tool.py
+++ b/core/cat/mad_hatter/decorators/tool.py
@@ -1,3 +1,5 @@
+import time
+import asyncio
 import inspect
 
 from typing import Union, Callable, List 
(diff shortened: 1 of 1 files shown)


pieroit:
@MaxDam thanks
@Pingdred can you take a look at this when you have time?
I would have this functions as `StrayCat` methods, since they require the LLM and could be useful from any hook/tool/form

PR #755, opened 2024-03-17
Add text splitter hook

zAlweNy26:
# Description

This PR aims to add a new hook to be able to customize the TextSplitter used by the RabbitHole when splitting chunks of text in different documents.

## Type of change

# Checklist:

### DIFF
--- a/core/cat/rabbit_hole.py
+++ b/core/cat/rabbit_hole.py
@@ -34,13 +34,25 @@ class RabbitHole:
             "text/markdown": TextParser(),
             "text/html": BS4HTMLParser()
         }
+        
+        self.__text_splitter = RecursiveCharacterTextSplitter.from_tiktoken_encoder(
+            separators = ["\\n\\n", "\n\n", ".\\n", ".\n", "\\n", "\n", " ", ""],
+            encoding_name = "cl100k_base",
+            keep_separator = True,
+            strip_whitespace = True
+        )
 
         self.__reload_file_handlers()
+        self.__reload_text_splitter()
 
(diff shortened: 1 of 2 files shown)


core/cat/rabbit_hole.py
```diff
@@ -401,13 +412,10 @@ def __split_text(self, stray, text, chunk_size, chunk_overlap):
...
         )
 
-        text_splitter = RecursiveCharacterTextSplitter.from_tiktoken_encoder(
-            chunk_size=chunk_size,
-            chunk_overlap=chunk_overlap,
-            separators=["\\n\\n", "\n\n", ".\\n", ".\n", "\\n", "\n", " ", ""],
-            encoding_name = "cl100k_base",
-            keep_separator=True,
-            strip_whitespace = True
+        text_splitter = self.text_splitter.from_tiktoken_encoder(
```
nicola-corbellini, line 415: what if one implements a splitter that doesn't have this method?

zAlweNy26: I checked the code, this method is by default in the TextSplitter and since we accept any subclass of it, it must have that method.

zAlweNy26: https://api.python.langchain.com/en/latest/base/langchain_text_splitters.base.TextSplitter.html#langchain_text_splitters.base.TextSplitter

nicola-corbellini: not sure this is the right strategy

pieroit:
Thanks @zAlweNy26 !
Made a few tweaks to give precedence `endpoint` > `hook`

PR #762, opened 2024-03-26
Add endpoint to message with Cat

zAlweNy26:
# Description

I also:
- fixed some typos around the codebase
- replaced old logic for user id

## Type of change

# Checklist:

### DIFF
--- a/core/cat/headers.py
+++ b/core/cat/headers.py
@@ -56,10 +56,10 @@ def check_api_key(request: Request, api_key: str = Security(api_key_header)) -> 
 
 
 # get or create session (StrayCat)
-def session(request: Request) -> str:
+def session(request: Request) -> StrayCat:
 
     strays = request.app.state.strays
-    user_id = request.headers.get("user_id")
+    user_id = request.headers.get("user_id", "user")
     event_loop = request.app.state.event_loop
     
     if user_id not in strays.keys():
--- a/core/cat/looking_glass/stray_cat.py
+++ b/core/cat/looking_glass/stray_cat.py
@@ -204,15 +204,17 @@ class StrayCat:
         if isinstance(self._llm, BaseChatModel):
             return self._llm.call_as_llm(prompt, callbacks=callbacks)
 
-    async def __call__(self, user_message_json):
+    async def __call__(self, message, save=True):
             """Call the Cat instance.
 
             This method is called on the user's message received from the client.
 
             Parameters
             ----------
-            user_message_json : dict
+            message : dict
                 Dictionary received from the Websocket client.
+            save : bool, optional
+                If True, the user's message is stored in the chat history. Default is True.
 
             Returns
             -------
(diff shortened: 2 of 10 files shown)


pieroit:
Thanks there are good ideas in the PR

Questions:
- why the `save` param? Looks a lot like the prompt settings we ended up taking away, and nobody asked for it
- what happens to tokens streaming and notifications when a client uses `POST /message`?

I agree on http endpoints, but what do we say to devs? Use http without streaming or use websocket for streaming?

What do you think @nicola-corbellini @Pingdred @valentimarco

zAlweNy26:
Hi @pieroit, thanks for the response.
- The `save` param is just a boolean param to decide whether or not to save the response in the chat history.
- Everything works fine, no notifications (because it's not a file upload) and no tokens streaming.

I don't know if HTTP streaming can be implemented, but I can check.
About the devs, just tell them that they have another way to message the cat, it's logical that an HTTP endpoint doesn't have streaming by default.

valentimarco:
> Thanks there are good ideas in the PR
> 
> Questions:
> 
>     * why the `save` param? Looks a lot like the prompt settings we ended up taking away, and nobody asked for it
> 
>     * what happens to tokens streaming and notifications when a client uses `POST /message`?
> 
> 
> I agree on http endpoints, but what do we say to devs? Use http without streaming and use websocket for streaming?
> 
> What do you think @nicola-corbellini @Pingdred @valentimarco

Save messages is a pretty good feature and can be good when for example we don't want the tool response (with redirect on true) in the chat history!

We should find a way to generate the messages via http like most of runners do..

nicola-corbellini:
I don't understand how one could pass `save=True` or `False`. I mean, since it is in the `__call__` method, I have no use cases in mind where one uses `stray_cat(message, save=True/False)`

zAlweNy26:
@nicola-corbellini I didn't mean "internal" use cases, but probably from the outside someone could want to get a cat response without saving the answer to chat history. Some use cases could be: privacy, testing, resources management.

zAlweNy26:
This makes even more sense if we finally decide to save chat history into the db

nicola-corbellini:
Exactly, I don't understand in which case, from the outside, one calls directly the instance like this `stray_cat()`

zAlweNy26:
Then I don't understand your doubt. What else should be called?

nicola-corbellini:
write a snippet in which you use the parameter please

zAlweNy26:
```js
fetch('/message?save=false', {
  method: "POST",
  headers: { "Content-Type": "application/json" },
  body: JSON.stringify({
    text: "hello cat!"
  }),
})
```

nicola-corbellini:
Ah lol now I get it, I didn't see it was a param in the new endpoint

zAlweNy26:
Buy a new pair of glasses

nicola-corbellini:
make two separate PRs for typos and new logic 😛

zAlweNy26:
No I won't

PR #788, opened 2024-04-24
"before_cat_stores_procedural_memory" hook created and "agent_allowed_tools" hook called

MaxDam:
added "before_cat_stores_procedural_memory" and "agent_allowed_tools" hooks.
Modified the following files:

"embed_procedures" method in cheshire_cat.py
"before_cat_stores_procedural_memory" method in flow.py
"execute_procedures_agent" method in agent_manager.py


pieroit:
Thanks @MaxDam !!
The hook `before_cat_stores_procedural_memory` is not aligned with rabbithole hooks, that pass arround langchain docs.
At the moment no solution came to my mind, let's see how people use them and if they get confused.

pieroit:
@MaxDam had to remove the `before_cat_stores_procedural_memory` hook while updating tests, too dangerous :(

Issue #690, opened 2024-01-26
Authenticators factory


pieroit:
An old design decision was to lock under AUTH_KEY only http endpoints.
Let's make the key cover *all* the endpoints, included websocket

EDIT: A better idea is to have AUTH_KEY for http endpoints and a different WS_AUTH_KEY for authenticating websocket

P.S. check if client libs need an update

diegogosmar:
I guess this is the python file requiring some modification in order to check the auth_key before allowing the websocket connection:
https://github.com/cheshire-cat-ai/core/blob/7a99ba2bcb33a95b241ab6944e6288f3a2a4950e/core/cat/routes/websocket.py#L43

pieroit:
Let's ask @EugenioPetulla where is better to force api auth because he handled the http one

Thanks

lucapirrone:
Hi guys, I'll give my opinion on this.
HTTP APIs and websocket communication play two different roles within the cat. HTTP APIs are used to administer the cat while websocket communication is used to let users communicate.

Using the same key for both the websocket and the http API would mean having to make the user administrator key public, causing a major security problem.

If you wanted to make the websocket secure, you should extend the cat with a plugin in order to implement a custom authorization logic based on your user management system.

For example, in an application with a token-based authentication system, you could insert the user token into the metadata of websocket messages and check the validity of the token in the "Before agent starts" hook.

What do you think about it?

diegogosmar:
Hi Luca,
I agree we should keep separate the HTTP RESTful API from the Websocket protocol.
However, I'm not sure we should have an external plugin to grant the Websocket security.

How about to add an APIKey related to the websocket initial authentication (different from the APIKey used for the Rest API)?
Ideally it would be to add the possibility of using an OAuth Token:
OAuth Tokens: For more robust security, especially in systems where users have individual accounts, consider using OAuth tokens. After the user logs in via a secure HTTP connection, your server issues a token that the client then presents when establishing the WebSocket connection.

I've also noticed that in the documentation examples the websockets are set not to use a secure connection (ws instead of wss):
config = ccat.Config(
    base_url=cat_url,
    port=1865,
    user_id="user",
    auth_key=auth_key,
    secure_connection=False  # Indicates an insecure connection (ws://)
)

I didn't try to set it to True and see if the WSS are working, however we should have a try to be sure that we support encrypted websocket communications.

Thanks!
ciao
Diego

pieroit:
[...]

PR #794, opened 2024-04-26
Authenticators Factory

giovannialbero1992:
# Description

Based on the discussion #690 

Related to issue #(issue)

## Type of change

# Checklist:

### DIFF
--- a/core/cat/headers.py
+++ b/core/cat/headers.py
@@ -1,7 +1,13 @@
 import os
 import fnmatch
 
-from fastapi import Request
+from typing import Annotated
+from cat.log import log
+
+from fastapi import (
+    WebSocket,
+    Request,
+)
 from fastapi import Security, HTTPException
 from fastapi.security.api_key import APIKeyHeader
 
@@ -18,8 +24,32 @@ The keys are piped with a `|`, hence the list takes care of splitting and storin
 
 api_key_header = APIKeyHeader(name="access_token", auto_error=False)
 
+def ws_auth(
+    websocket: WebSocket,
+    ) -> None | str:
+    """Authenticate endpoint.
 
-def check_api_key(request: Request, api_key: str = Security(api_key_header)) -> None | str:
+    Check the provided key is available in API keys list.
+
+    Parameters
+    ----------
+    request : Request
+        HTTP request.
+
+    Returns
+    -------
+    api_key : str | None
+        Returns the valid key if set in the `.env`, otherwise return None.
+
+    Raises
+    ------
+    HTTPException
+        Error with status code `403` if the provided key is not valid.
+
+    """
+    return websocket.app.state.ccat.authorizator.is_ws_allowed(websocket)
+    
+def http_auth(request: Request) -> None | str:
     """Authenticate endpoint.
 
     Check the provided key is available in API keys list.
@@ -42,16 +72,13 @@ def check_api_key(request: Request, api_key: str = Security(api_key_header)) -> 
         Error with status code `403` if the provided key is not valid.
 
     """
-    if not API_KEY:
+    authorizator = request.app.state.ccat.authorizator
+    if authorizator.is_master_key(request) or authorizator.is_http_allowed(request):
         return None
-    if fnmatch.fnmatch(request.url.path, "/admin*"):
-        return None
-    if api_key in API_KEY:
-        return api_key
     else:
         raise HTTPException(
             status_code=403,
-            detail={"error": "Invalid API Key"}
+            detail={"error": "Invalid Credentials"}
         )
 
 
--- a/core/cat/factory/authorizator.py
+++ b/core/cat/factory/authorizator.py
@@ -1,0 +1,69 @@
+from os import getenv
+from typing import Type
+from cat.log import log
+from pydantic import BaseModel, ConfigDict
+from cat.factory.custom_authorizator import BaseAuth, AuthorizatorNoAuth, AuthorizatorApiKey
+from cat.mad_hatter.mad_hatter import MadHatter
+
+class AuthorizatorSettings(BaseModel):
+    _pyclass: Type[BaseAuth] = None
+
+    @classmethod
+    def get_authorizator_from_config(cls, config):
+        if cls._pyclass is None or issubclass(cls._pyclass.default, BaseAuth) is False:
+            raise Exception(
+                "Authorizator configuration class has self._pyclass==None. Should be a valid Authorizator class"
+            )
+        return cls._pyclass.default(**config)
+
+class AuthorizatorNoAuthConfig(AuthorizatorSettings):
+    _pyclass: Type = AuthorizatorNoAuth
+
+    model_config = ConfigDict(
+        json_schema_extra={
+            "humanReadableName": "No Authorizator",
+            "description": "No authorizator is used. All requests are allowed.",
+            "link": "",
+        }
+    )
+
+class AuthorizatorApiKeyConfig(AuthorizatorSettings):
+    api_key: str
+    _pyclass: Type = AuthorizatorApiKey
+
+    model_config = ConfigDict(
+        json_schema_extra={
+            "humanReadableName": 'API Key Authorizator',
+            "description": 'Authoriza requests based on API key',
+            "link": "",
+        }
+    )
+
+def get_allowed_authorizator_strategies():
+    list_authorizator_default = [
+        AuthorizatorNoAuthConfig,
+        AuthorizatorApiKeyConfig,
+    ]
+
+    mad_hatter_instance = MadHatter()
+    list_authorizator = mad_hatter_instance.execute_hook(
+        "factory_allowed_authorizators", list_authorizator_default, cat=None
+    )
+
+    return list_authorizator
+
+def get_authorizators_schemas():
+    AUTHORIZATOR_SCHEMAS = {}
+    for config_class in get_allowed_authorizator_strategies():
+        schema = config_class.model_json_schema()
+        schema["auhrizatorName"] = schema["title"]
+        AUTHORIZATOR_SCHEMAS[schema["title"]] = schema
+    
+    return AUTHORIZATOR_SCHEMAS
+
+def get_authorizator_from_name(name):
+    list_authorizator = get_allowed_authorizator_strategies()
+    for authorizator in list_authorizator:
+        if authorizator.__name__ == name:
+            return authorizator
+    return None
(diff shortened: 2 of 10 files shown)


lucagobbi:
Hi @giovannialbero1992 I think you are doing a great job, this is really flexible and easy to implement. I propose also to have a generic abstract class from which custom authorizators should inherit, since `is_allowed` and `is_allowed_ws` methods should be the same for every implementation:

```python
class Auth(ABC):
    @abstractmethod
    def is_allowed(self, request: Request) -> bool:
        """
        Abstract method to validate if the HTTP request is allowed based on the implemented
        authentication and authorization mechanism.
        This method should return True if the request is allowed, or False otherwise.
        """
        pass

    @abstractmethod
    def is_allowed_ws(self, websocket: WebSocket) -> bool:
        """
        Abstract method to validate if the WebSocket connection is allowed based on the
        implemented authentication and authorization mechanism.
        This method should return True if the connection is allowed, or False otherwise.
        """
        pass
```

then the AuthSettings model should look something like this:

```python
class AuthSettings(BaseModel):
    _pyclass: Type[Auth] = None

    @classmethod
    def get_authorizator_from_config(cls, config: Dict):
        if cls._pyclass is None or not issubclass(cls._pyclass, Auth):
            raise ValueError(
                "Invalid configuration: _pyclass must be a subclass of Auth."
            )
        return cls._pyclass(**config)
```

I think this enforcement ensures consistency across different auth strategies, provides type safety on settings, reducing type errors and gives an easier way to devs to implement this.

pieroit:
> Hi @giovannialbero1992 I think you are doing a great job, this is really flexible and easy to implement. I propose also to have a generic abstract class from which custom authorizators should inherit, since `is_allowed` and `is_allowed_ws` methods should be the same for every implementation:
> 
> ```python
> class Auth(ABC):
>     @abstractmethod
>     def is_allowed(self, request: Request) -> bool:
>         """
>         Abstract method to validate if the HTTP request is allowed based on the implemented
>         authentication and authorization mechanism.
>         This method should return True if the request is allowed, or False otherwise.
>         """
>         pass
> 
>     @abstractmethod
>     def is_allowed_ws(self, websocket: WebSocket) -> bool:
>         """
>         Abstract method to validate if the WebSocket connection is allowed based on the
>         implemented authentication and authorization mechanism.
>         This method should return True if the connection is allowed, or False otherwise.
>         """
>         pass
> ```
> 
> then the AuthSettings model should look something like this:
> 
> ```python
> class AuthSettings(BaseModel):
>     _pyclass: Type[Auth] = None
> 
>     @classmethod
>     def get_authorizator_from_config(cls, config: Dict):
>         if cls._pyclass is None or not issubclass(cls._pyclass, Auth):
>             raise ValueError(
>                 "Invalid configuration: _pyclass must be a subclass of Auth."
>             )
>         return cls._pyclass(**config)
> ```
> 
> I think this enforcement ensures consistency across different auth strategies, provides type safety on settings, reducing type errors and gives an easier way to devs to implement this.

Agree on the parent class, also allows for more methods we may add later that may be overrideen or not from a subclass

P.S.: why ABS and not directly pydantic [BaseModel as we do for LLM/embedders](https://github.com/cheshire-cat-ai/core/blob/eb97809bba393ada05ef7b1c3a2ebbc3e2cc67f1/core/cat/factory/llm.py#L22)?
The latest covers having configs saved as dictionaries and calling _pyclass(**config), I don't how it works for ABC

lucagobbi:
> > Hi @giovannialbero1992 I think you are doing a great job, this is really flexible and easy to implement. I propose also to have a generic abstract class from which custom authorizators should inherit, since `is_allowed` and `is_allowed_ws` methods should be the same for every implementation:
> > ```python
> > class Auth(ABC):
> >     @abstractmethod
> >     def is_allowed(self, request: Request) -> bool:
> >         """
> >         Abstract method to validate if the HTTP request is allowed based on the implemented
> >         authentication and authorization mechanism.
> >         This method should return True if the request is allowed, or False otherwise.
> >         """
> >         pass
> > 
> >     @abstractmethod
> >     def is_allowed_ws(self, websocket: WebSocket) -> bool:
> >         """
> >         Abstract method to validate if the WebSocket connection is allowed based on the
> >         implemented authentication and authorization mechanism.
> >         This method should return True if the connection is allowed, or False otherwise.
> >         """
> >         pass
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > then the AuthSettings model should look something like this:
> > ```python
> > class AuthSettings(BaseModel):
> >     _pyclass: Type[Auth] = None
> > 
> >     @classmethod
> >     def get_authorizator_from_config(cls, config: Dict):
> >         if cls._pyclass is None or not issubclass(cls._pyclass, Auth):
> >             raise ValueError(
> >                 "Invalid configuration: _pyclass must be a subclass of Auth."
> >             )
> >         return cls._pyclass(**config)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think this enforcement ensures consistency across different auth strategies, provides type safety on settings, reducing type errors and gives an easier way to devs to implement this.
> 
> Agree on the parent class, also allows for more methods we may add later that may be overrideen or not from a subclass
> 
> P.S.: why ABS and not directly pydantic BaseModel as langchain does? The latest covers having configs saved as dictionaries and calling _pyclass(**config), I don't how it works for ABC - maybe BaseModel is not needed

If you are referring to AuthSettings I'd still keep the pydantic model, since it needs data validation on the auth class you choose, but the auth class imho should be a simple interface (i.e. an abstract class in python) with focus on behaviour rather than data validatio. Having a pydantic model here seems quite an overhead to me. Isn't it?

pieroit:
> If you are referring to AuthSettings I'd still keep the pydantic model, since it needs data validation on the auth class you choose, but the auth class imho should be a simple interface (i.e. an abstract class in python) with focus on behaviour rather than data validatio. Having a pydantic model here seems quite an overhead to me. Isn't it?

@lucagobbi fair enough!! Asking also to @Pingdred if you want to chip in

core/cat/headers.py
```diff
@@ -18,8 +24,32 @@
...
+        Returns the valid key if set in the `.env`, otherwise return None.
+
+    Raises
+    ------
+    HTTPException
+        Error with status code `403` if the provided key is not valid.
 
-def check_api_key(request: Request, api_key: str = Security(api_key_header)) -> None | str:
+    """
+    return websocket.app.state.ccat.authorizator.is_ws_allowed(websocket)
```
lucapirrone, line 50: In this case you still need to check the master_key as in the http request, otherwise the admin panel will break if the dev implements a custom rule in is_allowed_ws

giovannialbero1992: I'm not so sure about this, I try to expand a little bit my consideration.
If the idea is to keep backward compatible the actual behavior, with this solution the websocket remains open.
So setting an API_KEY any endpoint will be protected and the websocket not.
What do you think?
I involve even @zAlweNy26 that knows how the admin panel works.

lucapirrone: i don't agree with keeping websocket chat open, this is a terrible security issue. If a dev wants to implement custom security autorizer for the websocket connection, it should not be public but the admin still needs to work.
To allow that we always have to somehow pass the api key from the admin panel.

In the default authorizer "BaseAuth" the check on is_ws_allowed returns true so by default the old admin should works until the dev implement an authorizer with custom is_ws_allowed logic.

To avoid having backward compatibility problems we need to extend the admin panel to somehow pass the api key in websocket connection.

Little recap:
with currrent websocket check implementation -> authorizator.is_ws_allowed(websocket)
using BaseAuth:
-> **old admin panel** will works because is_ws_allowed returns true (✅) (No backward compatibility)
-> **direct websocket** connection should works without api key because is_ws_allowed returns true (✅)
using CustomAuth with custom is_ws_allowed logic
-> **admin panel** should not work because the admin panel doesn't know the authorization logic (❌)
-> **direct websocket** connection should works implementing custom dev authorizer logic (✅)

implementing this websocket check -> authorizator.is_ws_master_key(websocket) or authorizator.is_ws_allowed(websocket)
using BaseAuth:
-> **old admin panel** should still works because is_ws_allowed returns true (✅) (No backward compatibility)
-> **direct websocket** connection should works both with and without api key (✅)
using CustomAuth with custom is_ws_allowed logic
-> **old admin panel** should not work (Backward compatibility -> **we need to pass api key**) (❌)
-> **direct websocket** connection should works implementing custom dev authorizer logic (✅)


what do you think? @giovannialbero1992 @zAlweNy26

pieroit: I think we are overcomplicating because two things happen:
- the Cat is designed as a machine to machine microservice
- the admin is not machine to machine

We currently understand what is going on but this could be really hard to get for newbies.
As a long term vision, I'd love the Cat to be used standalone, and *also* as a microservice.
Most of all, it has to be all around easy to use :)

I take it as my own fault, for insisting so much on both having no users, and having the admin included in core.
I take it also on my shoulders to merge this PR and directly adding the following (help needed only on the admin):

## Admin user

There will be an user system into core, and as a start it will only include the `admin` user.
Credentials for the admin will be changeable via two env variables:
`CCAT_ADMIN_USER` // defaults to `admin`
`CCAT_ADMIN_PASSWORD` // defaults to `admin`
This is a raw start for a full fledged user system, that can evolve without breaking the admin panel.

When admin panel opens, it can do as follows:
- search for a JWT in localStorage, if not
- shows a username / password modal
- sends credentials to the core and receives a temp JWT

JWT is used for:
- http requests via `Authorization` header
- ws communication via a "token" key included in each message (or only on connection via URL parameter)

## Machine 2 machine

As it is now, `API_KEY` will cover the http endpoints and websocket stays open.
We could easily add a `WS_API_KEY` to also secure websockets.

In M2M communication, as far as I know, there is no need for a JWT and keys can be used directly.
Nothing prevents to use the user/pass credential system also for M2M.

## Summary

- Browsers are authenticated via credentials and JWT
- M2M is authenticated both like a browser or simply via static keys.
- Auth can be customized at will as we already designed, and the admin panel is saved.

Custom auth stays as @giovannialbero1992 implemented it, with minor modifications.
I'm going all in, tell me if you see something I'm not seeing.

pieroit:
Thank you @giovannialbero1992 and all the great minds involved in this.
I merged into develop and added some tests and refinements.

`CCAT_API_KEY` env variable for http is *always* checked when it is defined, both for retrocompatibility and for an additional layer of security. You can write your custom auth without worrying about that ;)
(see `BaseAuth._is_http_allowed`)

The admin panel auth will be on the side and will not impact the custom Auth.
Working on it the next days, also need to add more tests with a custom Auth in `mock_plugin`

Thaaaanks

PR #795, opened 2024-05-01
Skip `lost+found` directory

debu99:
# Description

Skip `lost+found` subdirectory in the plugin when mount volume for the plugin directory

Related to issue #(issue)

## Type of change

# Checklist:

### DIFF
--- a/core/cat/mad_hatter/mad_hatter.py
+++ b/core/cat/mad_hatter/mad_hatter.py
@@ -102,11 +102,17 @@ class MadHatter:
         log.info("ACTIVE PLUGINS:")
         log.info(self.active_plugins)
 
+        # Directories to skip
+        skip_directories = ['lost+found']
+
         # discover plugins, folder by folder
         for folder in all_plugin_folders:
-            self.load_plugin(folder)
-
             plugin_id = os.path.basename(os.path.normpath(folder))
+
+            if plugin_id in skip_directories:
+                continue
+
+            self.load_plugin(folder)
             
             if plugin_id in self.active_plugins:
                 self.plugins[plugin_id].activate()


pieroit:
When does this folder appear? Why is it a problem?
Thanks @debu99

pieroit:
Hi @debu99, better to not hardcode folder names relative to operating systems.

What do you think about turning the ERROR in a WARNING [here](https://github.com/cheshire-cat-ai/core/blob/3c675c2dddec75f608c8c9ba5af3b6f180319d09/core/cat/mad_hatter/mad_hatter.py#L127)?

Exception is already well handled, because the plugin is not created and the folder is skipped.


Thanks a lot!

core/cat/mad_hatter/mad_hatter.py
```diff
@@ -102,11 +102,17 @@ def find_plugins(self):
         log.info("ACTIVE PLUGINS:")
         log.info(self.active_plugins)
 
+        # Directories to skip
+        skip_directories = ['lost+found']
```
matteocacciola, line 106: what about skipping `__pycache__` as well?

Ilyasslamti:
**[Architecture Review] - Incomplete Blacklist Approach**

Hi @debu99, thanks for the contribution. However, hardcoding an exclusion for `lost+found` is a band-aid fix. Relying on a directory name blacklist is unscalable and leaves the plugin loader vulnerable to failing on other OS-generated or hidden directories (e.g., `.DS_Store`, `.git`, `.idea`).

**Required Enterprise-Grade Implementation:**
Instead of checking directory names, the loader must enforce **Structural Validation**. The system should verify the existence of a required manifest or entry point (e.g., `plugin.json` or the main `.py` entry file) *before* attempting to load any directory as a plugin. 

If a directory lacks this mandatory signature, the loader must silently bypass it. This implicitly handles `lost+found` and inherently future-proofs the system against any non-plugin directories. Please refactor this PR to use whitelist validation rather than name-based blacklisting.

PR #796, opened 2024-05-02
Insert "after_rabbithole_insert_memory" hook after document is stored

dedalo944:
Added "after_rabbithole_insert_memory" hook.
Files modified rabbit_hole.py, rabbithole.py


pieroit:
thanks @dedalo944 , will review soon ;)

pieroit:
Thanks @dedalo944, made a few improvements:
- changed the name to `after_rabbithole_stored_documents`
- hook recevices the points as stored by Qdrant, after they are inserted (so you have the idd and embeddings to retrieve them directly, plus content and metadata)
- the hook does not need (nor it is useful) to return anything, because they are already sotred.

Your use case should be satisfied, because you want to work on the vector db after insertion!
Let me know

PR #819, opened 2024-05-14
Fixed content attribute BUG in agent_manager

giovannaloro:
While trying to use the cat with Cohere API i found a bug in cat-->looking_glass-->agent_manager. Memory_chain.invoke returned a string but i was treated as an object.


Pingdred:
The type of response depends on the type of LLM used, chat or text generation; this fix breaks support for chat models (which we are currently working on develop). I am closing the PR, if you want to discuss other solutions let's talk on discord

valentimarco:
or try to use the update of develop

giovannaloro:
> or try to use the update of develop

what do you mean ?

pieroit:
> > or try to use the update of develop
> 
> what do you mean ?

He means, update your `develop` branch because the guys are fixing that

Issue #839, opened 2024-05-27
[Feature] WhiteRabbit expansion


jacopopalumbo01:
About the path to follow for the WhiteRabbit. During the dev meeting we discussed:
1. Add the WhiteRabbit inside the StrayCat.
2. Pass the StrayCat instance to the WhiteRabbit when calling any WhiteRabbit's function.
3. Additional methods (we should define in particular what to add).

About the first two points I have the following idea:
- The StrayCat has the WhiteRabbit instance
```python
...
# The Stray cat goes around tools and hook, making troubles
class StrayCat:
    """User/session based object containing working memory and a few utility pointers"""

    def __init__(
            self,
            user_id: str,
            main_loop,
            ws: WebSocket = None,
        ):
        self.__user_id = user_id
        self.working_memory = WorkingMemory()
        self.white_rabbit = WhiteRabbit()
...
```
- The WhiteRabbit has all methods with a StrayCat instance parameter (as done in the actual code for `schedule_chat_message`.
- The StrayCat "wraps" the WhiteRabbit functions to make the cat instance "invisible" to the plugin dev: (example with `schedule_cat_message`)
```python
        ...
        # Inside StrayCat

        def schedule_chat_message(self, content: str, days=0, hours=0, minutes=0, seconds=0, milliseconds=0, microseconds=0):
                self.white_rabbit.schedule_chat_message(content=content, cat=self, days=days, hours=hours, minutes=minutes, seconds=seconds, milliseconds=milliseconds, microseconds=microseconds)
         ...
```
By doing that the plugin developer can schedule a message just like that:
```python
from cat.mad_hatter.decorators import hook

@hook
def before_cat_sends_message(final_output, cat):    
    cat.schedule_chat_message(content="White Rabbit: Oh dear! Oh dear! I shall be too late!", minutes=1)
    return final_output
```
It's a little bit a continuous communication between the StrayCat and the WhiteRabbit but it should make very easy the integration inside the plugins. @pieroit @lucagobbi what do you think?

lucagobbi:
Hi Jacopo, thanks for organizing this discussion.

Regarding point 1):

I think we should stick with the current code and make WhiteRabbit a property of StrayCat just like MadHatter, RabbitHole, etc. it makes the code more consistent

```python
...
@property
def white_rabbit(self):
    return CheshireCat().white_rabbit
...
```

Regarding point 2) and your proposal:

[...]

PR #840, opened 2024-05-28
Scheduler update

jacopopalumbo01:
# Description

Now the StrayCat has the WhiteRabbit property. Added the `schedule_jobs`. There are some missing things (like the other types of scheduling discussed in #839.

With this update you can schedule a job like follows:
```python
def job(cat):
    cat.send_ws_message(content="White Rabbit: Oh dear! Oh dear! I shall be too late!", msg_type="chat")
    
@hook
def before_cat_sends_message(final_output, cat):    
    cat.white_rabbit.schedule_job(job, seconds=10, cat=cat)
    return final_output
```

Notice: If you don't need the cat instance in the scheduled job you can just ignore it, now it is handled as a kwarg. This means that you can also pass other attributes to the method. Example:
```python
def job(log_message):
    log.debug(log_message)
    
@hook
def before_cat_sends_message(final_output, cat):    
    cat.white_rabbit.schedule_job(job, seconds=10, log_message="Hello world")
    return final_output
```

Related to PR #836 discussion and #839 

## Type of change

# Checklist:

### DIFF
--- a/core/cat/looking_glass/stray_cat.py
+++ b/core/cat/looking_glass/stray_cat.py
@@ -544,6 +544,10 @@ Allowed classes are:
     @property
     def agent_manager(self):
         return CheshireCat().agent_manager
+    
+    @property
+    def white_rabbit(self):
+        return CheshireCat().white_rabbit
 
     @property
     def loop(self):
--- a/core/cat/looking_glass/white_rabbit.py
+++ b/core/cat/looking_glass/white_rabbit.py
@@ -102,6 +102,38 @@ class WhiteRabbit:
         # Schedule the job
         self.scheduler.add_job(job, 'date', run_date=schedule, kwargs=kwargs)
         
+    def schedule_interval_job(self, job, start_date:datetime = None, end_date:datetime = None, days=0, hours=0, minutes=0, seconds=0, **kwargs):
+        """
+        Schedule an interval job
+
+        Parameters
+        ----------
+        job: function
+            The function to be called.
+        start_date: datetime
+            Start date. If None the job can start instantaneously
+        end_date: datetime
+            End date. If None the job never ends.
+        days: int
+            Days to wait.
+        hours: int
+            Hours to wait.
+        minutes: int
+            Minutes to wait.
+        seconds: int
+            Seconds to wait.
+        **kwargs
+            The arguments to pass to the function
+        """
+        
+        # Check that the function is callable
+        if not callable(job):
+            log.error("WhiteRabbit: The job should be callable!")
+            raise TypeError(f"TypeError: '{type(job)}' object is not callable")
+
+        # Schedule the job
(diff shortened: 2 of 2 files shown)


pieroit:
Looks good to me!

The StrayCat instance passing is now present both in RabbitHole and WhiteRabbite, at one point we could figure out a way to implicitly pass it, whenever the methods are called from a Stray instance. I'll propose for Cat v2, because it requires architectural changes :D

image

jacopopalumbo01:
Update:
- Added a `listener` to the scheduler. Now there is a log about the scheduled job behavior after job ends.
- Added `schedule_interval_job()` to schedule recurrent jobs

Next thing to do: implement `cron` type.

Another interesting idea could be a decorator for jobs that do not start inside a hook or a tool.
It should be something like that:
```python
@schedule(date=date)
def func():
     log.info("WhiteRabbit: Hi there!")
```
What do you think?

P.S.
APScheduler job can also accept an id or a name as parameter. Do you think that we should add those arguments? I'm thinking about managing jobs. We can provide a method to get the scheduled jobs and one to delete them. By having custom ids and names it could be more accessible.

lucagobbi:
Really like this! About the `@schedule` annotation I'm quite neutral, one could instantiate the CheshireCat or directly the WhiteRabbit in his/here plugin code, since they are singleton and schedule the function directly there. But, if it is a requested feature it's for sure less verbose with the annotation.

About the id/name of the scheduled jobs I totally agree: if you start a recurrent task at the moment you cannot delete it at runtime (maybe inserting some logic in the scheduled job itself, but sometimes may not be such a cool idea). Plus, in the future, one could build a plugin(the  _WhiteRabbitInAction_) that creates job, deletes them, pauses them, etc.

jacopopalumbo01:
Update: Added ids management, `get_jobs`, `remove_job`, `schedule_cron_job` methods.

pieroit:
> Update:
> 
> * Added a `listener` to the scheduler. Now there is a log about the scheduled job behavior after job ends.
> * Added `schedule_interval_job()` to schedule recurrent jobs
> 
> Next thing to do: implement `cron` type.
> 
> Another interesting idea could be a decorator for jobs that do not start inside a hook or a tool. It should be something like that:
> 
> ```python
> @schedule(date=date)
> def func():
>      log.info("WhiteRabbit: Hi there!")
> ```
> 
> What do you think?
> 
> P.S. APScheduler job can also accept an id or a name as parameter. Do you think that we should add those arguments? I'm thinking about managing jobs. We can provide a method to get the scheduled jobs and one to delete them. By having custom ids and names it could be more accessible.

Awesome!
Let's wait for a dedicated decorator, after there is some usage for this utility we can better understand what is needed

Thanks @jacopopalumbo01 @lucagobbi

Issue #818, opened 2024-05-13
[Refactor] Abstract Vector Memory and simple api from plugins


valentimarco:
1. Abstract API for the Vectory Memory (so in future we are able to attach different dbs)
2. Refactor the embedding stuff (@pieroit i don't remember why we need it, wasn't for the tool's examples?)
3. Add api for doing stuff in plugins, for example 'cat.declerative_memory.filter({})'

pieroit:
> 1. Abstract API for the Vectory Memory (so in future we are able to attach different dbs)

Not strictly necessary to be honest, we fly well with Qdrant. Abstracting is a good idea to encapsulate it and avoid breaking changes coming from them

> 2. Refactor the embedding stuff (@pieroit i don't remember why we need it, wasn't for the tool's examples?)

`VectorMemoryCollection` class has become spaghetti code, it is not on par with other ones

> 3. Add api for doing stuff in plugins, for example 'cat.declerative_memory.filter({})'

Nice!

valentimarco:
in the straycat:
```py
def recall(query: str | Embedding, metadata: dict  = none, search_declerative: bool = true, search_procedural: bool = true, search_episodic: bool = true, override_working_memory: bool = true): Set[Document]
```
Method to search in memory and must have:
- query > a string or an embedding
- metadata > to filter stuff 
-  some booleans to do stuff (maybe use **kargs to do that stuff can be clean)

nicola-corbellini:
Maybe it can be useful to return also the scores and the embeddings of the recalled points. 
E.g., at first, it could return `Set[Tuple[str, float, List[float]]]` like there is in other parts of the framework.
Probably I'm saying something trivial, but, later on, we could start designing a custom `Memory` class like 
```python
class Memory
    text: str
    embedding: List[float]
    metadata: dict
    score: float (?)
    def to_document(self): -> Document
```
also, other params for the `recall` method may be: `min_score: float` to recall memories with a minimum score, `k: int` for the max number of memories returned and if `k=-1|None` the method could just use `scroll` from qdrant and return all the points.

valentimarco:
I like the `Memory` class but should be uniform in all framework

pieroit:
Awesome, agree with @nicola-corbellini on having a custom class to throw memories/points/documents around.
Step 1: Follow the same output as `StrayCat.recall_memories_to_working_memory`
Step 2: standardize with internal class

About the method, to reelaborate on @valentimarco 's proposal

```python
[...]

PR #848, opened 2024-06-07
add `recall` to stray_cat.py

nicola-corbellini:
# Description

This is a draft of the `recall` method discussed in #818.
I'm leaving some questions in a self-review.

Related to issue #(issue)

## Type of change

# Checklist:


core/cat/looking_glass/stray_cat.py
```diff
@@ -156,6 +157,40 @@ def send_error(self, error: Union[str, Exception]):
...
+        else:
+            collections = self.memory.vectors.collections.keys()
+
+        for collection in collections:
+            vector_memory = getattr(self.memory.vectors, collection)
+            if k:
+                memories = vector_memory.recall_memories_from_embedding(
+                    query, metadata, k, threshold
+                )
+                if override_working_memory:
```
nicola-corbellini, line 188: overriding is made possible only if the dev defines a `k`. Hence, he or she is not scrolling the whole collection.
Is this ok?

core/cat/looking_glass/stray_cat.py
```diff
@@ -156,6 +157,40 @@ def send_error(self, error: Union[str, Exception]):
...
+                    memory_key = f"{collection}_memories"
+                    setattr(self.working_memory, memory_key, memories)
+            else:
+                memories = vector_memory.get_all_points()
```
nicola-corbellini, line 192: Yet the method as no return value because I wanted to ask how you intended to merge the memories of the different collections.
E.g. should we return a `dict` like
```python
memories = {
    "episodic_memories": # results of the recall from episodic,
   "declarative_memories": # results of the recall from declarative
}
```
similar to what we save in the working memory? Or maybe, do we want to have a rerank strategy and merge everything in a single collection?
An hybrid (and dirty?) solution could be to have an additional parameter to choose between reranking and separate results

nicola-corbellini (commented):
Hi guys, here a couple of questions for you @pieroit @valentimarco

core/cat/looking_glass/stray_cat.py
pieroit: - if `override_working_memory == True`, just override what is contained in working memory
- if `override_working_memory == False`, do not override working memory
- in both cases, return the dictionary you prototyped above, with contents in the same format as we do for working memory

core/cat/looking_glass/stray_cat.py
```diff
@@ -17,7 +18,7 @@
 from cat.convo.messages import CatMessage, UserMessage, MessageWhy, Role
 
 MSG_TYPES = Literal["notification", "chat", "error", "chat_token"]
-
+MEMORY_COLLECTION = Literal["episodic", "declarative", "procedural"]
```
pieroit, line 21: No need for this, when you ask for core collections:

```python
collections = self.memory.vectors.collections.keys()
```

You can then loop over them and see if the requested ones are present.
If collection is not present, ignore and throw a warning

core/cat/looking_glass/stray_cat.py
pieroit: Additional improvements if you get the time:
- method `recall_relevant_memories_to_working_memory` uses internally `recall` with `override_working_memory == True` and filters as defined by hooks
- add tests for the method in `tests/looking_glass/test_stray_cat.py` talking to the stray without the endpoints (example [here](https://github.com/cheshire-cat-ai/core/blob/31cc1ee4456c9a54cefe3c9812d433dd77f4206c/core/tests/looking_glass/test_stray_cat.py#L10))

pieroit (changes requested):

nicola-corbellini:
eehy @pieroit sorry for the late response. I don't think i will be able to be back on this task any time soon, maybe @nickprock would you like to address the changes? i remember you showed interested in developing the method

PR #852, opened 2024-06-10
add get_job, pause_job and resume_job methods plus minor adjustments

lucagobbi:
# Description

Hi! As discussed in the last dev meeting, I propose adding some methods to the `WhiteRabbit` class to finalize the excellent work by @jacopopalumbo01 and provide a complete and useful API wrapping APScheduler features. The added methods are:

 - `get_job`: Returns a dictionary of a single scheduled job.
 - `pause_job`: Temporarily pauses a scheduled job.
 - `resume_job`: Resumes a paused job.

Additionally, I have made some minor adjustments for type hinting and corrected comment typos. I have replaced every `id` reference with `job_id` to avoid shadowing built-in names. However, if we consider this a breaking change, we can restore `id`.

Regarding other features, I think we are covering most use cases, but APScheduler also provides for a concatenation of triggers, which could be really interesting in situations where triggers are complex:

```python

from apscheduler.triggers.combining import OrTrigger
from apscheduler.triggers.cron import CronTrigger

# run at 10:00 from Monday to Friday, but also at 11:00 from Saturday to Sunday

trigger = OrTrigger(
    CronTrigger(day_of_week="mon-fri", hour=10),
    CronTrigger(day_of_week="sat-sun", hour=11),
)

```

Currently we need to access the inner scheduler to achieve something like that, it would be nice to provide a simple API also in the WhiteRabbit class. 

Related to issue #839 

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

### DIFF
--- a/core/cat/looking_glass/white_rabbit.py
+++ b/core/cat/looking_glass/white_rabbit.py
@@ -1,3 +1,5 @@
+from typing import Dict, List
+
 from pytz import utc
 from datetime import datetime, timedelta
 
@@ -22,7 +24,7 @@ class WhiteRabbit:
     def __init__(self):
         log.info("Initializing WhiteRabbit...")
         
-        # Where the jobs are stored. We can also use an external db to have persistency
+        # Where the jobs are stored. We can also use an external db to have persistence
         jobstores = {
             'default': MemoryJobStore()
         }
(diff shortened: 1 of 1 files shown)


jacopopalumbo01:
Looks perfect to me! About the concatenation of triggers we should definitely explore them. However they are an "advanced" feature of WhiteRabbit so maybe it could be better to wait and see if the WhiteRabbit is used by the users as we discussed yesterday in the dev meeting.

pieroit:
Awesome :trophy:

PR #856, opened 2024-06-14
Fix WhiteRabbit (it was always late)

jacopopalumbo01:
# Description

WhiteRabbit was always late because the container's timezone wasn't the same as the host machine. This update fixes the problem without any change from the user perspective if using Linux or MacOS.
However it's a little bit tricky on Windows. User should set an environment variable before `docker-compose up` like follows:
```bash
$ set TZ=Europe/Rome
```
where `Europe/Rome` is chosen from this [list](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones#List).

To make it easier maybe we can create a bash script for windows to detect automatically the local timezone.
## Type of change

# Checklist:

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -29,3 +29,6 @@
 
 # CONFIG_FILE
 # CCAT_METADATA_FILE="cat/data/metadata.json"
+
+# Set container timezone
+# CCAT_TIMEZONE=Europe/Rome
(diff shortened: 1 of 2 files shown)


pieroit:
Thanks @jacopopalumbo01 
Does seem overkill to open volumes just for the timezone, let's have everything as as the same env variable `CCAT_TIMEZONE`

Inside core we should deal as much as possible only with unix timestamps

jacopopalumbo01:
@pieroit I've updated the PR. Now it only uses `CCAT_TIMEZONE`.

pieroit:
Danke !

PR #862, opened 2024-06-22
Replace equals with levenshtein method

lucagobbi:
# Description

As discussed in issue #859, we should avoid strict equality checks for classified strings and instead use a fuzzy approach like the Levenshtein distance. To prevent outliers from being forced into the nearest label, a threshold is set at `0.5`. This is fairly permissive but should be suitable for most use cases. 

Let's try it out

## Type of change

# Checklist:


pieroit:
Looks reasonable
I'll test thanks @lucagobbi !

Issue #864, opened 2024-06-26
Store intermediate prompts, replies, and token count for each message - and send back to the client


pieroit:
For every message submitted to the Cat, we can store in working_memory:
- tokens used (input and output)
- prompts used
- replies for each prompt

Info can be sent back to the client in the `why`
This would allow for easier debugging and better estimates on resources usage.

Just a proposal: interactions could be accumulated in a list, something like:

```python

# init (before cat reads message)
cat.working_memory.model_interactions = []

# at each LLM usage
cat.working_memory.model_interactions.append(
  ModelInteraction(
    model_type="llm"
    source="ProceduresAgent"
    prompt="some prompt",
    reply="llm output",
    input_tokens=340,
    output_tokens=100
  )
)

# at each embedder usage
cat.working_memory.model_interactions.append(
  ModelInteraction(
    model_type="embedder"
    source="recall"
    prompt="some prompt",
    reply=[0.3, 0.1, 0.87],
    input_tokens=340,
    output_tokens=0
  )
)

# when seding response back to client
CatMessage.why.model_interactions = cat.working_memory.model_interactions
```

# Token count

We could  count how many tokens have been used for an interaction, *if* the count is general for all LLMs.
*No OpenAI only solutions!*

## Input tokens
For input tokens, we could use `tiktoken` which is already a dependency in core.
Problem is, tiktoken is OpenAI only, so we can count the tokens in a prompt but not sure the estimate will hold for non-OpenAI LLMs.
Probably not much of  a difference?

## Output otkens
For outptut tokens, add a langchain callback*, which would be independent of the vendor, but only work for streaming LLMs.
Or count with tiktoken once the llm reply is back.

---
*We already have a callback [`NewTokenHandler`](https://github.com/cheshire-cat-ai/core/blob/31cc1ee4456c9a54cefe3c9812d433dd77f4206c/core/cat/looking_glass/callbacks.py#L5), we can add `TokenCounter` (so count is disjoint from sending each token to the client)

zAlweNy26:
Amazing idea 💯

PR #867, opened 2024-07-06
Track model interactions

lucagobbi:
# Description

This PR aims to prepare the groundwork for tracking model interactions and resource usage. 

@pieroit I believe this feature is a valuable addition to the framework because, especially for newcomers, the Cat is often treated as a black box, logs can sometimes be challenging to interpret. Providing a comprehensive observability of what is really happening under the hood can signficantly improve the dev ex with the framework.

However, Im still facing some ancestral doubts:

- Should `ModelInteraction` be a unique class? Since the embedder model interaction type has no output tokens, a fixed source (`recall`, is it the only one?) and a fixed type for reply `List[float]`, maybe it should be better to split `ModelInteraction` into two separated classes `LLMModelInteraction` and `EmbedderModelInteraction`?
- Another central issue is the standardized tokenizer. Following @pieroit 
suggestions in the issue,  I used `tiktoken`, the OpenAI tokenizer, for both input tokens and output tokens (to avoid excluding non-streaming APIs). I've stumbled upon different discussions online and it seems that the community is facing an error of `±15%` depending on the provider (here a [link](https://github.com/gkamradt/LLMTest_NeedleInAHaystack/issues/25) to an interesting discussion about that). I feel like we're taking a very rigid approach by using tiktoken everywhere, I've experimented also with response metadata but Langchain is still generic about [that](https://api.python.langchain.com/en/latest/outputs/langchain_core.outputs.llm_result.LLMResult.html#langchain_core.outputs.llm_result.LLMResult) and providers are not standardizing their responses.

Tested with OpenAI (gpt-4o, gpt-3.5-turbo), Cohere (command), Ollama (gemma2:9b).

Related to issue #864 

## Type of change

# Checklist:

### DIFF
--- a/core/cat/looking_glass/callbacks.py
+++ b/core/cat/looking_glass/callbacks.py
@@ -1,5 +1,9 @@
+from typing import Any, Dict, List
 from langchain.callbacks.base import BaseCallbackHandler
-
+from langchain_core.outputs.llm_result import LLMResult
+import tiktoken
+from cat.log import log
+from cat.convo.messages import ModelInteraction
 
 class NewTokenHandler(BaseCallbackHandler):
     def __init__(self, stray):
@@ -8,3 +12,38 @@ class NewTokenHandler(BaseCallbackHandler):
 
     def on_llm_new_token(self, token: str, **kwargs) -> None:
         self.stray.send_ws_message(token, msg_type="chat_token")
+
+
+class ModelInteractionHandler(BaseCallbackHandler):
+    """
+    Langchain callback handler for tracking model interactions.
+    """
+
+    def __init__(self, stray, source: str):
+        self.stray = stray
+        self.stray.working_memory.model_interactions.append(
+            ModelInteraction(
+                model_type="llm",
+                source=source,
+                prompt="",
+                reply="",
+                input_tokens=0,
+                output_tokens=0,
+            )
+        )
+
+
+    def on_llm_start(self, serialized: Dict[str, Any], prompts: List[str], **kwargs) -> None:
+        encoding = tiktoken.get_encoding("cl100k_base")
+        input_tokens = sum(len(encoding.encode(prompt)) for prompt in prompts)
+        self.last_interaction.prompt = ''.join(prompts)
+        self.last_interaction.input_tokens = input_tokens
+
+    def on_llm_new_token(self, token: str, **kwargs) -> None:
+        self.last_interaction.output_tokens += 1
+
+    def on_llm_end(self, response: LLMResult, **kwargs) -> None:
+        self.last_interaction.reply = response.generations[0][0].text
+    @property
+    def last_interaction(self) -> ModelInteraction:
+        return self.stray.working_memory.model_interactions[-1]
--- a/core/cat/looking_glass/callbacks.py
+++ b/core/cat/looking_glass/callbacks.py
@@ -2,7 +2,9 @@ from typing import Any, Dict, List
 from langchain.callbacks.base import BaseCallbackHandler
 from langchain_core.outputs.llm_result import LLMResult
 import tiktoken
-from cat.convo.messages import ModelInteraction
+from cat.convo.messages import LLMModelInteraction
+import time
+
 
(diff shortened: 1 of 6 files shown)


core/cat/looking_glass/callbacks.py
```diff
@@ -8,3 +11,41 @@ def __init__(self, stray):
...
+                source=source,
+                prompt="",
+                reply="",
+                input_tokens=0,
+                output_tokens=0,
+            )
+        )
+
+    def _count_tokens(self, text: str) -> int:
+        # cl100k_base is the most common encoding for OpenAI models such as GPT-3.5, GPT-4 - what about other providers?
```
pieroit, line 35: Awesome, let's stick with this for the moment!

pieroit:
Hi @lucagobbi thanks for this, awesome feat

> @pieroit I believe this feature is a valuable addition to the framework because, especially for newcomers, the Cat is often treated as a black box, logs can sometimes be challenging to interpret. Providing a comprehensive observability of what is really happening under the hood can signficantly improve the dev ex with the framework.

Agree, also noobs will require less support from us

> * Should `ModelInteraction` be a unique class? Since the embedder model interaction type has no output tokens, a fixed source (`recall`, is it the only one?) and a fixed type for reply `List[float]`, maybe it should be better to split `ModelInteraction` into two separated classes `LLMModelInteraction` and `EmbedderModelInteraction`?

Your call!
I guess it can be a good idea. What we care about is an easy interpretation of what happened in the backend, so the focus is on devx.
Another pro of separating is to have a `report` including both calls to models and calls to other parts of the framework (hooks, tools, vector memory - which can be many). Can we generalize? A list of interactions, all subclassing a main class.
I suggest in the main class we also add a timestamp

> * Another central issue is the standardized tokenizer. Following @pieroit
>   suggestions in the issue,  I used `tiktoken`, the OpenAI tokenizer, for both input tokens and output tokens (to avoid excluding non-streaming APIs). I've stumbled upon different discussions online and it seems that the community is facing an error of `±15%` depending on the provider (here a [link](https://github.com/gkamradt/LLMTest_NeedleInAHaystack/issues/25) to an interesting discussion about that). I feel like we're taking a very rigid approach by using tiktoken everywhere, I've experimented also with response metadata but Langchain is still generic about [that](https://api.python.langchain.com/en/latest/outputs/langchain_core.outputs.llm_result.LLMResult.html#langchain_core.outputs.llm_result.LLMResult) and providers are not standardizing their responses.

@lucagobbi @valentimarco @zAlweNy26 
On the admin report we can show `~3200 tokens` to indicate that it is an estimation.
Most people needs this to make calculations on vendor spend for a typical interaction, so I would not worry much about the estimate being exact, at least until langchain itself offers the information

lucagobbi:
I've updated the PR splitting the `ModelInteraction` class into:

- `LLMModelInteraction` class 
- `EmbedderModelInteraction` class

Base `ModelInteraction` class has now a `started_at` timestamp, while LLMModelInteraction has also an `ended_at` timestamp to measure response duration (is it useful?)

@pieroit sorry I dont actually get what you mean by `report`, is this moving in the right direction?

pieroit:
> I've updated the PR splitting the `ModelInteraction` class into:
> 
> * `LLMModelInteraction` class
> * `EmbedderModelInteraction` class
> 
> Base `ModelInteraction` class has now a `started_at` timestamp, while LLMModelInteraction has also an `ended_at` timestamp to measure response duration (is it useful?)

Awesome thanks @lucagobbi !

> 
> @pieroit sorry I dont actually get what you mean by `report`, is this moving in the right direction?

Yes, I mean instead of only model interactions, we log in the list also other stuff (tools, hooks, vector memory queries).
This can be the next version of the `why`.
I can merge this for the moment and we later discuss how to include other info.

pieroit:
So sweet 

Screenshot from 2024-07-16 12-45-39

zAlweNy26:
I'll upgrade the Admin panel to show this :)

pieroit:
> I'll upgrade the Admin panel to show this :)

Propose a wireframe bedore coding please

Issue #850, opened 2024-06-10
[BUG] Entering admin with https generates many errors


rafleze:
I deployed a cat instance on a server where kubernetes orchestrates the service. I added the CORE_USE_SECURE_PROTOCOLS=true  parameter but when I navigate the admin I have many errors.

![Screenshot 2024-06-10 alle 10 53 10](https://github.com/cheshire-cat-ai/core/assets/1647787/86f866ab-77dc-4ea6-ad93-973ee7cda7f3)

the problem is that in the endpoint called without the slash. when it redirects it breaks because it does it in http

![image](https://github.com/cheshire-cat-ai/core/assets/1647787/922d4dc7-31f9-4e81-b532-4d16f8089981)

Maybe the solution is explained at this link: https://stackoverflow.com/questions/63511413/fastapi-redirection-for-trailing-slash-returns-non-ssl-link

At the moment I solved the problem adding this parameter: FORWARDED_ALLOW_IPS

But I suppose that the redirect could be avoided adding "/" at the end of the url.

zAlweNy26:
Can you try this PR #837 and tell us if the problem persists?

pieroit:
Should be already fixed in `develop`

PR #875, opened 2024-07-25
add: new env variables to manage https proxy pass

kodaline:
## Problem

- setup the cat with an https nginx proxy
- when using `/rabbithole` endopint, or any endopint that makes a redirect, fastapi/uvicorn will downgrade the connection to `http` because of the missing uvicorn server flags
- in case the original connection is https this will trigger the following error:

```
Mixed Content
The page at was loaded over HTTPS but requested an insecure resource
This request has been blocked the content must be served over HTTPS
```

## Proposed solution
- add new environment variable `CCAT_HTTPS_PROXY_MODE` in order to add the required flags when needed

## Reference
https://github.com/tiangolo/fastapi/discussions/9328#discussioncomment-8242245

## Type of change

# Checklist:

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -6,6 +6,12 @@
 # Decide to use https / wss secure protocols
 # CCAT_CORE_USE_SECURE_PROTOCOLS=true
 
+# Uvicorn and FastAPI operating behind https proxy
+# CCAT_HTTPS_PROXY_MODE=true
+
+# Comma separated list of IPs to trust with proxy headers. A wildcard '*' means always trust.
+# CCAT_CORS_FORWARDED_ALLOW_IPS="*"
+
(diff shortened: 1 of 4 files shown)


kodaline:
> Would be great to have `forwarded_allow_ips` configurable! We can call it `CCAT_CORS_FORWARDED_IPS`
> 

Yep that is indeed a nice idea, I will separate the two uvicorn flags on two env variables :)

> Great job btw <3

Thanks <3

pieroit:
:heart_eyes_cat: 

I guess this fixes #850 ?

kodaline:
> 😻
> 
> I guess this fixes #850 ?

If the cat is correctly set up I guess it fixes also that case yep

Issue #895, opened 2024-08-26
[BUG] Cat keeps talking to strangers even if WS key is set


lucagobbi:
**Describe the bug**
I was testing the new auth implementation (btw great work @pieroit, @lucapirrone) and i found this little bug. If you lock the Cat installation by setting both api keys (`CCAT_API_KEY`, `CCAT_API_KEY_WS`) it works fine and requests throw a 403 if not authenticated. But if one decides to lock only WebSocket by using the `CCAT_API_KEY_WS` you can still talk to the Cat.

**To Reproduce**
Steps to reproduce the behavior (as example`):
1. Set only the `CCAT_API_KEY_WS` in your .env file
2. Restart the Cat
3. Start a Ws connection without passing any key or token

**Expected behavior**
WS should be locked even if only the WS key is set and not both.

PR #904, opened 2024-09-03
Cat won't talk to strangers if not allowed to

lucagobbi:
# Description

Currently, if you set only the CCAT_WS_API_KEY you'll still be able to talk to the Cat even when no key is passed. 

I fixed it refactoring the `authorize_user_from_key` method. 

Actually, I think that, in the future, we should move out the control: "if no keys are set, then always pass", but is it strictly related to the AuthHandler? If so, then its right to keep it there even if out of scope. Plus, is it correct to gave base permissions if no keys are set? Probably would be better to distinguish even there the requested resource.

Related to issue #895 

## Type of change

# Checklist:

### DIFF
--- a/core/cat/factory/custom_auth_handler.py
+++ b/core/cat/factory/custom_auth_handler.py
@@ -1,5 +1,6 @@
 from abc import ABC, abstractmethod
 from datetime import datetime, timedelta
+from typing import Literal
 from pytz import utc
 import jwt
 
@@ -21,6 +22,7 @@ class BaseAuthHandler(ABC):  # TODOAUTH: pydantic model?
 
     async def authorize_user_from_credential(
         self,
+        protocol: Literal["http", "websocket"],
         credential: str,
         auth_resource: AuthResource,
         auth_permission: AuthPermission,
@@ -36,7 +38,7 @@ class BaseAuthHandler(ABC):  # TODOAUTH: pydantic model?
         else:
             # API_KEY auth
             return await self.authorize_user_from_key(
-                user_id, credential, auth_resource, auth_permission
+                protocol, user_id, credential, auth_resource, auth_permission
             )
 
     @abstractmethod
@@ -49,6 +51,7 @@ class BaseAuthHandler(ABC):  # TODOAUTH: pydantic model?
     @abstractmethod
     async def authorize_user_from_key(
         self,
+        protocol: Literal["http", "websocket"],
         user_id: str,
         api_key: str,
         auth_resource: AuthResource,
@@ -94,6 +97,7 @@ class CoreAuthHandler(BaseAuthHandler):
 
     async def authorize_user_from_key(
             self,
+            protocol: Literal["http", "websocket"],
             user_id: str,
             api_key: str,
             auth_resource: AuthResource,
@@ -106,10 +110,10 @@ class CoreAuthHandler(BaseAuthHandler):
             return AuthUserInfo(
                 id=user_id,
                 name=user_id,
-                permissions=get_base_permissions()
+                permissions=get_full_permissions()
             )
 
-        if auth_resource == AuthResource.CONVERSATION:
+        if protocol == "websocket":
             return self._authorize_websocket_key(user_id, api_key, ws_key)
         else:
             return self._authorize_http_key(user_id, api_key, http_key)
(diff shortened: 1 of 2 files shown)


core/cat/factory/custom_auth_handler.py
```diff
@@ -93,37 +93,53 @@ async def authorize_user_from_jwt(
...
-        # TODOAUTH: should we consider the user_id or just give
-        #    admin permissions to all users with the right api keys?
-
-        # chatting over websocket
-        if auth_resource == AuthResource.CONVERSATION and api_key == ws_api_key:
+        if not http_key and not ws_key:
             return AuthUserInfo(
                 id=user_id,
                 name=user_id,
                 permissions=get_base_permissions()
```
pieroit, line 109: If no keys are set, I guess we can give full permissions to anybody?
In this way if there are no keys set, you cannot use most endpoints (I need to check)

core/cat/factory/custom_auth_handler.py
```diff
@@ -93,37 +93,53 @@ async def authorize_user_from_jwt(
...
             )
 
-        # any http endpoint
-        if api_key == http_api_key:
+        if auth_resource == AuthResource.CONVERSATION:
+            return self._authorize_websocket_key(user_id, api_key, ws_key)
```
pieroit, line 113: There are also endpoints for conversation not related to websocket :(

pieroit:
Thanks @lucagobbi I left some comments
Maybe we can have a chat to redesign a few things

core/cat/factory/custom_auth_handler.py
lucagobbi: Damn, you're right! So, currently, you have to pass the WS api key for the http message endpoint if only the WS key is set? And when only the http key is set if you not passing the api key you can message freely because None == None...

core/cat/factory/custom_auth_handler.py
lucagobbi: Yep, agree with that, otherwise we could differentiate by requested resource:

```python

if not http_key and not ws_key:
            return AuthUserInfo(
                id=user_id,
                name=user_id,
                permissions=get_base_permissions() if auth_resource == AuthResource.CONVERSATION else get_full_permissions()

```

lucagobbi:
> Thanks @lucagobbi I left some comments Maybe we can have a chat to redesign a few things

Thank you Piero, I think we might need some more adjustments, ping me whenever you want

pieroit:
Key / JWT permissions rationale:
- no CCAT_API_KEY, no CCATA_API_KEY_WS = full permissions on everything
- no CCAT_API_KEY, yes CCATA_API_KEY_WS = full permissions on http, ws only if key is present
- yes CCAT_API_KEY, no CCATA_API_KEY_WS = full permissions on http if you provide key, open ws
- yes CCAT_API_KEY, yes CCATA_API_KEY_WS = full permissions on both if you provide the key

JWT:
- we assume JWT is used *when key are set*

P.S.: do not enter AuthHandler if no key is set [(here)](https://github.com/cheshire-cat-ai/core/blob/f14f2c0acebddd3f1f1efb6ba3b161bea45e9e3a/core/cat/auth/connection.py#L40)

lucagobbi:
> API KEY permissions rationale:
> 
> * no CCAT_API_KEY, no CCATA_API_KEY_WS = full permissions on everything
> * no CCAT_API_KEY, yes CCATA_API_KEY_WS = full permissions on http, ws only if key is present
> * yes CCAT_API_KEY, no CCATA_API_KEY_WS = full permissions on http if you provide key, open ws
> * yes CCAT_API_KEY, yes CCATA_API_KEY_WS = full permissions on both if you provide the key
> 
> JWT:
> 
> * if you are using JWT without setting the keys????
> * in this case the auth handlers are active
> 
> P.S.: intervention point [(here)](https://github.com/cheshire-cat-ai/core/blob/f14f2c0acebddd3f1f1efb6ba3b161bea45e9e3a/core/cat/auth/connection.py#L40)

I cant manage to find a way to do what we have said during the last dev meeting without a proper refactor of those classes, which, for the moment, I'd avoid. 

my alt proposal is to:

- set full_permissions if no keys are set (as you've suggested in your first comment)
- add a specific auth resource for the HTTP message endpoint like: `AuthResource.MESSAGE` or distinguish two conversation resources: `AuthResource.CONVERSATION_WS` / `AuthResource.CONVERSATION_HTTP`. In this way we can keep this PR as is, following the API KEY rationale you proposed, fixing those little security bugs and keeping separate ws controls from http ones.

What do you think?

lucagobbi:
@pieroit I've added the protocol based control that we talked about yesterday, this decouples auth key methods from auth resources. What do you think about it?

pieroit:
Looks great! Will check asap locally

pieroit:
Thanks @lucagobbi for the contribution and patience :)
I rewrote the tests on keys to check them independently

PR #911, opened 2024-09-14
Replace deprecated configuration type python with debugpy

sambarza:
# Description

VSCode launch configuration type "python" is deprecated, replaced with "debugpy"


pieroit:
I'll try soon thanks

Issue #909, opened 2024-09-13
Lower default threshold for recall


pieroit:
New embedder models (especially multilingual) are way more finegraned and conservative than embedders from a few months ago.

If a threshold of 0.7/0.8 was reasonable then, it is not now and it should be way lower.

It is possible to change thresholds via plugin (e.g. Cat Advanced Tools), but the default ones should also be lowered.

Intervention point: [here](https://github.com/cheshire-cat-ai/core/blob/f14f2c0acebddd3f1f1efb6ba3b161bea45e9e3a/core/cat/looking_glass/stray_cat.py#L211)

scicco:
Hello, I've seen the tag 'good first issue' and I tried to start contributing to this feature.

pieroit:
Merging soon

PR #912, opened 2024-09-16
lowering recall threshold

scicco:
# Description

This is my first attempt to try to contribute a bit to this awesome project😹 
This commit will lower the recall threshold. It also separates the config introducing the `RecallSettings` class as suggested in [TODO](https://github.com/cheshire-cat-ai/core/blob/f14f2c0acebddd3f1f1efb6ba3b161bea45e9e3a/core/cat/looking_glass/stray_cat.py#L212)

I've set the following constant values:
```
    DEFAULT_K = 3
    DEFAULT_TRESHOLD = 0.5
```
please let me know if those values are appropriate.

Related to issue (#909)

## Type of change

# Checklist:

### DIFF
--- a/core/cat/looking_glass/recall_settings.py
+++ b/core/cat/looking_glass/recall_settings.py
@@ -1,0 +1,31 @@
+"""Module for retrieving default configurations for episodic, declarative and procedural memories"""
+
+
+class RecallSettings:
+    """Class for retrieving default configurations for episodic, declarative and procedural memories"""
+
+    DEFAULT_K = 3
+    DEFAULT_TRESHOLD = 0.5
+
+    def _build_settings(
+        self,
+        recall_query_embedding,
+        user_id=None,
+        k=DEFAULT_K,
+        treshold=DEFAULT_TRESHOLD,
+    ):
+        return {
+            "embedding": recall_query_embedding,
+            "k": k,
+            "threshold": treshold,
+            "metadata": {"source": user_id} if user_id else None,
+        }
+
+    def default_episodic_config(self, recall_query_embedding, user_id):
+        return self._build_settings(recall_query_embedding, user_id)
+
+    def default_declarative_config(self, recall_query_embedding):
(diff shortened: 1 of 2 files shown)


valentimarco:
Thx for the contribution! For me is ok but i would prefer to have a central module to store all classes like this one. What do you think @pieroit ?

core/cat/looking_glass/recall_settings.py
```diff
@@ -0,0 +1,31 @@
+"""Module for retrieving default configurations for episodic, declarative and procedural memories"""
+
+
+class RecallSettings:
```
pieroit, line 4: Hi @scicco having a class is a good idea, but this one is a little over engineered imho.
Try to have a simple class extending `cat.utils.BaseModelDict` so it can be used both as an object (leveraging a pydantic models with types and defaults) and as a dict and will be retrocompatible.
Examples [here](https://github.com/cheshire-cat-ai/core/blob/f14f2c0acebddd3f1f1efb6ba3b161bea45e9e3a/core/cat/convo/messages.py#L39)

```python
class RecallSettings(BaseModelDict):
    embedding
    k
    threshold
    # skipping types and defaults because I'm cooked
    ...
```

Then when it is time to use it:

```python
default_episodic_recall_config = RecallSettings(
    embedding=recall_query_embedding,
    user_id=self.user_id
)
```

pieroit (changes requested):

pieroit:
@scicco sorry for the dely, left some suggestions. Let me know if you can do it or I'll try

pieroit (changes requested):
left some comments

scicco:
No problem, I'll work on that

core/cat/looking_glass/recall_settings.py
scicco: done

scicco:
Hello @pieroit, I've implemented the changes. If something should be changed or fixed, let me know, and I'll do it.

Thank you

PR #914, opened 2024-09-17
WebSocket fix proposal

fatualux:
# Description

This is a proposal to fix [this issue](https://github.com/cheshire-cat-ai/core/issues/915)

#### FIX 1

The first step is to edit **stray_cat.py**:

Basically, we add a try-except block to the original code (`__send_ws_json`) to handle the problem:

```python
from websockets.exceptions import ConnectionClosedOK

    def __send_ws_json(self, data: Any):
        try:
            # Run the corutine in the main event loop in the main thread
            # and wait for the result
            asyncio.run_coroutine_threadsafe(
                self.__ws.send_json(data), loop=self.__main_loop
            ).result()

        except ConnectionClosedOK as ex:
            if ex.code == 1000:
                log.warning(ex)
                if self.__ws:
                    del self.__ws
                    self.__ws = None
```

It catches the *ConnectionClosedOK* exception, and, If the connection is closed with the code 1000 (normal closure), it logs a warning using *log.warning(ex)*.

Then, if the `ConnectionClosedOK` exception is caught, it deletes the `self.__ws` variable and sets it to `None`.

In the screencast below we can observe how the fix works against [reproduce.py](https://github.com/user-attachments/files/17024336/reproducer.md):

fix1

### FIX 2

The first fix is used to handle this specific case: a given user closes the connection while the Cat is still providing an answer.

In case of a hypothetical different flow, we would run into the same problem again.

To avoid this, we can generalize the fix, adding this try-except block to the async method `get_user_stray` in connection.py:

```python
    if stray._StrayCat__ws:
    try: 
        await stray._StrayCat__ws.close()
    except RuntimeError as ex:
        log.warning(ex)
        if stray._StrayCat__ws:
            del stray._StrayCat__ws
            stray._StrayCat__ws = None
```

This fix ensure that, if we get a RuntimeError from WebSocket, we get a warning, and the *stray*, that once again is an inconsistent state, is cleaned up (same strategy we used for the first fix).

To avoid using a private attribute, we could extract this and implement a new method in the `StrayCat` class:

- **stray_cat.py**

```python
    async def close_connection(self):
        if self.__ws:
            try:
                await self.__ws.close()
            except RuntimeError as ex:
                log.warning(ex)
                if self.__ws:
                    del self.__ws
                    self.__ws = None
```

- **connection.py**

```python
        if user.id in strays.keys():
            stray = strays[user.id]
            await stray.close_connection()
```

In the screencast below, again, we can see how the second fix works against [reproduce.py](https://github.com/user-attachments/files/17024336/reproducer.md):

fix2

Related to [this issue](https://github.com/cheshire-cat-ai/core/issues/915)

## Type of change

# Checklist:

I did not commented my code, I think it is not needed in this case.

As always, thanks for your work and support/review.


lucapiccinelli:
Hello, any news about this fix proposal? Would you like any help in reproducing the error? Or do you think the fix needs some changes? 
We provided a Python reproducer, but the error can also be reproduced by closing the Cat web UI while the cat is providing an answer.
We are available and open to any discussion.


Thank you 😊

pieroit:
> Hello, any news about this fix proposal? Would you like any help in reproducing the error? Or do you think the fix needs some changes? We provided a Python reproducer, but the error can also be reproduced by closing the Cat web UI while the cat is providing an answer. We are available and open to any discussion.
> 
> Thank you 😊

Hey @lucapiccinelli thanks for reaching out. In a few days I will check, test and merge, sorry for the delay, the PR is *greatly* appreciated

lucapiccinelli:
Ok thank you. No pressure, we worked around it as a temporary fix. I just wanted to understand of it ok, or there was any blocker

pieroit:
Please @fatualux redo the PR branching off `develop`, not off `main`.
Also check @sambarza suggestion in issue #915

After that we're done! Thanks :D

fatualux:
OK, I will do so, thanks for the review.

Issue #797, opened 2024-05-03
Endpoint and button for "factory reset"


pieroit:
Many people messes up with the embedders and LLMs, breaking the installation. Typical error comes from Qdrant:
`TypeError: '>=' not supported between instances of 'ValueError' and 'int'`

Let's have an endpoint to reset default llm and embedder (plugins should not be an issue).
Reset button could also be placed in admin settings for people not able to check out the terminal.

alfredotoselli:
Hello! A couple question about the endpoint:
- "reset default llm and embedder" means to set the dumb `Default Language Model` as LLM and the `Default Embedder` as Embedder? In this case the reset buttons in the admin panel (now seem inactive) will hit this new endpoint?
- from the user's perspective the concept of "factory reset" in practice means: reset llm; reset embedder; wipe all collections; right ?

![llm](https://github.com/user-attachments/assets/63c195fb-4800-4aa6-9569-fdc91ccb9e80)
![embedder](https://github.com/user-attachments/assets/31278b8c-dedc-4bfc-bd5b-4ba992d56c62)

pieroit:
> * "reset default llm and embedder" means to set the dumb `Default Language Model` as LLM and the `Default Embedder` as Embedder? In this case the reset buttons in the admin panel (now seem inactive) will hit this new endpoint?

Those reset button should only revert to the latest db config, if I remember correctly (@zAlweNy26 can you confirm?)

> * from the user's perspective the concept of "factory reset" in practice means: reset llm -> reset embedder -> wipe all collections right ?

It has become more complicated because there are now users and various settings.
The easiest way to reset is to delete `cat/data/metadata.json` and wipe all the memories.
Not sure how this endpoint should behave at the moment :/

zAlweNy26:
I don't remember that, I think it should reset to what it is when you first install the cat.

PR #916, opened 2024-09-17
feat: factory reset endpoint

alfredotoselli:
# Description

Hello! Happy to learn about the project and start contributing on technical stuff :)
Let's check if the endpoint makes sense as is and if any corner case comes to your minds. 
If ok I'll add a couple of tests before merging.

The situation will be:
- Possibility to wipe out the collections also from admin panel button
- Inactive reset buttons for llm and embedder on the settings panel
  - As mentioned in the issue we should decide if restore the last setting present on DB or the setting on the first installation
-  This new `/settings/factory-reset` endpoint (currently without button) that wipes out all the collections, deletes `metadata.json` and performs a "soft restart"

Related to issue #797 

## Type of change

# Checklist:


core/cat/routes/memory/collections.py
```diff
@@ -40,19 +41,14 @@ async def wipe_collections(
     """Delete and create all collections"""
 
     vector_memory: VectorMemory = stray.memory.vectors
-    collections = list(vector_memory.collections.keys())
+    deleted_memories = utils.delete_collections(vector_memory)
```
matteocacciola, line 44: if you move the `delete_collection` in the `vector_memory_collection.py` like
```
    def wipe(self) -> bool:
        try:
            self.client.delete(collection_name=self.collection_name)
            return True
        except Exception as e:
            log.error(f"Error deleting collection {self.collection_name}: {e}")
            return False
```

you could use

```
to_return = {str(c): ccat.memory.vectors.collections[str(c)].wipe() for c in MemoryCollection}
```

core/cat/routes/settings.py
```diff
@@ -102,3 +108,41 @@ def delete_setting(
...
+
+            os.remove(metadata_file)
+
+        vector_memory: VectorMemory = stray.memory.vectors
+        deleted_memories = utils.delete_collections(vector_memory)
+
+        utils.singleton.instances.clear()
+        request.app.state.strays.clear()
+
+        request.app.state.ccat = CheshireCat()
```
matteocacciola, line 139: does it work with a singleton, i.e. a class which constructor is triggered the first time only?

core/cat/routes/memory/collections.py
alfredotoselli: Thanks for the suggestion! I made slightly different thing because there is direct access to the memory from the stray

Issue #921, opened 2024-09-24
[Refactor] Refactoring `/message` endpoint in order to execute the agent


valentimarco:
At the moment, `/message` is mostly bugged when trying to interact with the agent.

The last dev meeting, we discuss how can be:
1. No streaming (If you need it, use WS.). The reason is simple: This http method will be design for agent interaction that are not conversational but rather utility (Translations, Interaction with CRMs etc...)
2. JSON that contains at least the key "text" (like we do in WS). This way if you have some custom stuff that reads the json, you are free to do it even in http.
3. Some python types to help OpenApi gen (PLS)

MihirKohli:
Can you please brief more about the issue.

mastrogiovanni:
What to do with websocket notifications? 
I fixed the /message endpoint with the following (mutuated from ws interface):
```python
@router.post("/message", response_model=CatMessage)
async def message_with_cat(
    payload: Dict = Body({"text": "hello!"}),
    stray=Depends(HTTPAuth(AuthResource.CONVERSATION, AuthPermission.WRITE)),
) -> Dict:
    """Get a response from the Cat"""
    payload["user_id"] = stray.user_id
    return await stray.__call__(payload)
```

The problem is that stray cat try to send notification to websocket producing a lot of warnings.
Is that ok if I'll remove the warnings and I'll leave the current behaviour that simply ignore the notifications if ws is None?

valentimarco:
Fixed with #932

PR #932, opened 2024-10-03
Run stray in threadpool to allow tool execution on http message endpoint

lucagobbi:
# Description

After a debug session with @valentimarco  we found that synchronous tools were not executed by the Stray if called via HTTP message endpoint, due to a problem with the event loop.

@Pingdred to the rescue, proposed to run the stray using FastAPI method `run_in_threadpool`.

Since the `_arun` method (in the procedural agent) may involve a mix of sync and async functions, using `run_in_threadpool` ensure that any potential blocking code is executed in a thread pool, avoiding event loop blockin. 

Am I right @Pingdred, @valentimarco?

Related to issue #921 

## Type of change

# Checklist:

- [] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/core/cat/looking_glass/stray_cat.py
+++ b/core/cat/looking_glass/stray_cat.py
@@ -450,17 +450,22 @@ class StrayCat:
 
         return final_output
 
-    def run(self, user_message_json):
+    def run(self, user_message_json, return_message=False):
         try:
             cat_message = self.loop.run_until_complete(self.__call__(user_message_json))
-            # send message back to client
-            self.send_chat_message(cat_message)
+            if return_message:
+                # return the message for HTTP usage
+                return cat_message
+            else:
+                # send message back to client via WS
+                self.send_chat_message(cat_message)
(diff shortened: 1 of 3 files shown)


valentimarco:
Correct!

pieroit:
Really delicate.
Thanks for checking this @lucagobbi @valentimarco @Pingdred 

In time we can tests for sync / async tools and the /message endpoint

lucagobbi:
Hola all, after our discussion I've updated the PR to hide the "loop" complexity in the `stray.run` method. Now `stray.run` is the only method to call the stray both via HTTP and WebSocket. It accepts a flag to decide whether to return directly the result or to send it over websocket.

Personally, I don't like it very much, but as we said we hide that complexity and make the run method the only the entrypoint to execute the stray.

valentimarco:
it's time to write some tests

pieroit:
Thanks @lucagobbi encapsulation is good

Issue #915, opened 2024-09-17
WebSocket inconsistent closing when a chat is in progress


fatualux:
### Description of the bug

**SUMMARY**

```python
    RuntimeError: Unexpected ASGI message 'websocket.close', after sending 'websocket close'.
```

When a user closes his *StrayCat* while a process is still running, a runtime error is raised, and the user can no longer interact with the Cat until the Cat is restarted.

**IN DEPTH**

When we connect to the Cat with a given client, a new connection (*Stray Cat*) to a WebSocket's server is created.

Let's say this user closes the connection while a process is still active (e.g. the Cat's is still answering to a query).

<ins>Inside the same session</ins> of the CheshireCat, if we connect with that user again, the Cat uses the same session; but now, that session has an object `websocket` <ins>in an inconsistent state</ins>.

This probably happens because the Cat's WebSocket object, a higher level implementation of the native Python WebSocket object (see [Starlette](https://www.starlette.io/) framework), has a status `Connected`, but actually the connection is closed (and the lower level Python WebSocket object has a status `Disconnected`).

### Steps to Reproduce

My collegue, [Luca](https://github.com/lucapiccinelli), implemented a script in order to reproduce this unexpected behaviour, systematically, getting an error message in the Cat's console.

**Requirements**:
- Python
- Virtualenv
- [reproduce.py](
[reproducer.md](https://github.com/user-attachments/files/17024596/reproducer.md) (I cannot attach python files, so I provide it in a markdown format)

In order to make this happen, we can follow these steps:

1. Open the Cat's core directory and start the Cat:
```bash
    docker compose up
```
2. Move ***reproduce.py*** in a separate folder, create a Python virtual environment and install the required dependency:

```bash
    mkdir reproducer && mv reproduce.py reproducer/
    cd reproducer && virtualenv .venv
    source .venv/bin/activate && pip install cheshire-cat-api
```

3. Launch the script, giving, at first:

```bash
    python reproduce.py 5 Alice
```

**Note**: `5` is the *sleep time* in seconds, `"Alice"` is the username passed to the function.

The second time we run the script, we change the sleep time to be <ins>less than one second</ins>:

```bash
    python reproduce.py 0.5 Alice
```

This time, we get this error:

```python
    RuntimeError: Unexpected ASGI message 'websocket.close', after sending 'websocket.close'.
```

From now on, we will no longer be able to connect to the WebSocket with the user `Alice`.
[...]

PR #936, opened 2024-10-07
WebSocket Fix Proposal

fatualux:
# Description

This is a proposal to fix [this issue](https://github.com/cheshire-cat-ai/core/issues/915).

#### FIX 1

The first step is to edit **stray_cat.py**:

Basically, we add a try-except block to the original code (`run` method) to handle the problem:

```python
from websockets.exceptions import ConnectionClosedOK

    def run(self, user_message_json, return_message=False):
        try:
            cat_message = self.loop.run_until_complete(self.__call__(user_message_json))
            if return_message:
                # return the message for HTTP usage
                return cat_message
            else:
                # send message back to client via WS
                self.send_chat_message(cat_message)
        except Exception as e:
            log.error(e)
            traceback.print_exc()
            if return_message:
                return {"error": str(e)}
            else:
                try:
                    self.send_error(e)
                except ConnectionClosedOK as ex:
                    log.warning(ex)
                    if self.__ws:
                        del self.__ws
                        self.__ws = None
```

It catches the *ConnectionClosedOK* exception, and, If the connection is closed with the code 1000 (normal closure), it logs a warning using *log.warning(ex)*.

Then, if the `ConnectionClosedOK` exception is caught, it deletes the `self.__ws` variable and sets it to `None`.

In the screencast below we can observe how the fix works against [reproduce.py](https://github.com/user-attachments/files/17024336/reproducer.md):

fix1

### FIX 2

The first fix is used to handle this specific case: a given user closes the connection while the Cat is still providing an answer.

In case of a hypothetical different flow, we would run into the same problem again.

To avoid this, we can generalize the fix, adding this try-except block to the async method `get_user_stray` in connection.py:

```python
    if stray._StrayCat__ws:
    try: 
        await stray._StrayCat__ws.close()
    except RuntimeError as ex:
        log.warning(ex)
        if stray._StrayCat__ws:
            del stray._StrayCat__ws
            stray._StrayCat__ws = None
```

This fix ensure that, if we get a RuntimeError from WebSocket, we get a warning, and the *stray*, that once again is an inconsistent state, is cleaned up (same strategy we used for the first fix).

To avoid using a private attribute, we could extract this and implement a new method in the `StrayCat` class:

- **stray_cat.py**

```python
    async def close_connection(self):
        if self.__ws:
            try:
                await self.__ws.close()
            except RuntimeError as ex:
                log.warning(ex)
                if self.__ws:
                    del self.__ws
                    self.__ws = None
```

- **connection.py**

```python
        if user.id in strays.keys():
            stray = strays[user.id]
            await stray.close_connection()
```

In the screencast below, again, we can see how the second fix works against [reproduce.py](https://github.com/user-attachments/files/17024336/reproducer.md):

fix2

Related to [this issue](https://github.com/cheshire-cat-ai/core/issues/915)

## Type of change

# Checklist:

I did not commented my code, I think it is not needed in this case.

As always, thanks for your work and support/review.

### DIFF
--- a/core/cat/auth/connection.py
+++ b/core/cat/auth/connection.py
@@ -140,11 +140,10 @@ class WebSocketAuth(ConnectionAuth):
 
         if user.id in strays.keys():
             stray = strays[user.id]
-            # Close previus ws connection
-            if stray._StrayCat__ws:
-                await stray._StrayCat__ws.close()
+            await stray.close_connection()
+
             # Set new ws connection
-            stray._StrayCat__ws = connection
+            stray.reset_connection(connection)
             log.info(
                 f"New websocket connection for user '{user.id}', the old one has been closed."
             )
--- a/core/cat/looking_glass/stray_cat.py
+++ b/core/cat/looking_glass/stray_cat.py
@@ -19,6 +19,7 @@ from cat.memory.working_memory import WorkingMemory
 from cat.convo.messages import CatMessage, UserMessage, MessageWhy, Role, EmbedderModelInteraction
 from cat.agents import AgentOutput
 from cat import utils
+from websockets.exceptions import ConnectionClosedOK
 
 MSG_TYPES = Literal["notification", "chat", "error", "chat_token"]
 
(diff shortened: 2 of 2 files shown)


pieroit:
Thanks @fatualux !!!
Welcome

PR #940, opened 2024-10-10
Do not start the container manually during tests action

sambarza:
# Description

We start the container just to run the tests however the launched instance is not used by the tests, FastAPI's `TestClient` doesn't need the Cat to be running; it starts the Cat instance on its own for each individual test.
Removed the up and down steps.

This should also speed up the tests step too.

Using `docker exec` instead of `docker run` because `exec` needs an already running container, while `run` creates the container before starting the command. `docker run` need the service name instead of the container name.

Not related to issue.

## Type of change

# Checklist:

- [X ] I have performed a self-review of my own code

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -22,9 +22,5 @@ jobs:
     runs-on: "ubuntu-latest"
     steps:
       - uses: actions/checkout@v2
-      - name: Cat up
-        run: docker compose up -d
       - name: Test
-        run: docker exec cheshire_cat_core python -m pytest --color=yes .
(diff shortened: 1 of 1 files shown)


pieroit:
Awesome, also for climate change :D

pieroit:
You are right there is no need to launch again for the tests.
Also tried this out locally, works with and without the dev instance on.

Just added a `--rm` flag to remove the ocntainer once the tests are done:

```bash
docker compose run --rm cheshire-cat-core python -m pytest --color=yes .
```

Opening an issue also on the docs

Issue #868, opened 2024-07-09
`fast_reply` hook


pieroit:
Hook `agent_fast_reply` is useful after recall, to plug a custom agent.

I propose an hook `fast_reply` which shortcircuits both recall and agent.
This is useful to:
- canned replies
- send stuff to the Cat via websocket when a reply is not really needed
- launch custom LLM chains / agents / pipelines (i.e. launching directly the LLM with a simple prompt, which is very much requested)

lucagobbi:
That's cool! I'm making some experiments out of it, two questions arised:

1. What about conversation history both in working and in episodic memory? Should we keep? I think we should, even if someone could want to avoid certain message to pass, hence to be stored in memory as well... That should be hookable as well?
2. What about the `why` response? Knowing that the `fast_reply` hook was triggered and it shortcircuited indeed the response is important in this case. That's what we were discussing in #864

pieroit:
> That's cool! I'm making some experiments out of it, two questions arised:
> 
> 1. What about conversation history both in working and in episodic memory? Should we keep it? I think we should, even if someone could want to avoid certain messages to pass, hence to be stored in memory as well... That should be hookable as well?

Agree let's keep it, later on if it is requested we may add a boolean to the websocket message to let the client decide what gets stored and what not
@zAlweNy26 proposed something similar to this few months ago

> 2. What about the `why` response? Knowing that the `fast_reply` hook was triggered and it shortcircuited indeed the response is important in this case. That's what we were discussing in [Store intermediate prompts, replies, and token count for each message - and send back to the client #864](https://github.com/cheshire-cat-ai/core/issues/864)

I propose to leave the why as is, if in `fast_reply` the dev calls recall or stores stuff in working_memory collections, the why will be populated.
Am I missing something?

Thanks!!!

pieroit:
After dev meeting 23/09/2024

![image](https://github.com/user-attachments/assets/ca944770-9f1f-47d2-b992-1fdcd80f0066)

- `fast_reply` jumps over any side effects* and must return a `CatMessage` (or dictionary with `output` key)
- `agent_fast_reply` jumps over just the agent, the framework takes care of side effects*. Must return a `AgentOutout` (or dictionary with `output` key)

* working memory updates (user message, convo history, episodic collection, model_interactions)

alfredotoselli:
I could take this

pieroit:
Merging soon thanks @alfredotoselli

PR #949, opened 2024-10-18
feat: fast_reply hook

alfredotoselli:
# Description

An attempt to tackle the related issue

- The new hook skips the memory recall and the main agent
- The user message that triggers the hook is still added to the episodic memory
- Conversation history is updated

---
- The `why` is not populated, not sure how to handle it in this case
- Not clear to me what "introduce proper cleaning of working memory" means

Related to issue #868 

## Type of change

# Checklist:

### DIFF
--- a/core/cat/agents/main_agent.py
+++ b/core/cat/agents/main_agent.py
@@ -47,14 +47,13 @@ class MainAgent(BaseAgent):
         stray.working_memory.agent_input = agent_input
 
         # should we run the default agents?
-        fast_reply = {}
-        fast_reply = self.mad_hatter.execute_hook(
-            "agent_fast_reply", fast_reply, cat=stray
+        agent_fast_reply = self.mad_hatter.execute_hook(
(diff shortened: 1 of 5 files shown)


pieroit:
Thanks @alfredotoselli will review soon

pieroit:
Hi @alfredotoselli thanks and sorry for the delay.

PR was quite good, a few notes:
- `_handle_fast_reply_message` method was not necessary, if a plugin uses `fast_reply` there should be *NO* side effects, which means working memory and episodic memory are *not* impacted.
- updated the tests accordingly. I found delightful the way you injected an hook inside the mad hatter; we should reason about having a folder in the tests dedicated to hook testing (all of them) using this methodology.

alfredotoselli:
Thanks for the feedback! Next iteration I can setup a folder for hook testing with this style.

Didn't get the first point, `_handle_fast_reply_message` is still there  so currently working and episodic memory ARE impacted

Issue #945, opened 2024-10-17
Replace sync request and I/O libs with async


alfredotoselli:
@lucagobbi 

Check `registry.py`

lucagobbi:
As mentioned in the PR we have a bunch of blocking code here and there that should be migrated to async. However. we must be aware that async is not always good both for readability and for what the code does itself, putting async in front of any function definition is not a solution at all. 

Additionally we need to keep in mind how FastAPI handle these kind of stuff, because if we declare a route as async it will run it in the main event loop and if we insert blocking code in there it freeze the entire server. At that point we have two path ahead:
1. Declare our route as normal def without async so that FastAPI will run that in a threadpool.
2. Remove the blocking code and make the function really async (e.g. not using `requests` for http calls but use an async client like `httpx` or `aiohttp`)


Here some [docs](https://fastapi.tiangolo.com/async/#very-technical-details) about it, and [here](https://github.com/fastapi/fastapi/discussions/6946) a discussion in which they state that approach 2 performs better

The same happens with background tasks, all our ingestion pipeline is based on that feature. We should think about it as well. [Here](https://stackoverflow.com/questions/67599119/fastapi-asynchronous-background-tasks-blocks-other-requests) some discussions about this.

Hope you find these resources useful and we can further discuss this issue

pieroit:
Let's stay on the threadpool until we have an actual measure of the worthyness of all this async.

Please give priority to devx (both ours and to the plugin developers)
If people want to scale, they can scale horizontally, the Cat is mostly for individuals and SME

PR #951, opened 2024-10-20
[WIP] To async or not to async

lucagobbi:
# Description

This PR removes blocking code from async route definitions, which is currently blocking the event loop. Async code is usually introduced to improve performance in IO bound scenarios like network calls, reading/writing files, etc. We have a bunch of those cases throughout the entire codebase, this is just a starting point.

It's important to note that not all code should be made asynchronous for the sake of readability and practicality, we should be mindful of how synchronous and asynchronous code are handled within our FastAPI application, 'cause simply putting async before any method is not a solution.

Related to issue #945 

## Type of change

# Checklist:

### DIFF
--- a/core/cat/mad_hatter/registry.py
+++ b/core/cat/mad_hatter/registry.py
@@ -62,7 +62,7 @@ async def registry_search_plugins(
         return []
 
 
-async def registry_download_plugin(url: str) -> str:
+async def registry_download_plugin(url: str):
     log.info(f"Downloading {url}")
 
     registry_url = get_registry_url()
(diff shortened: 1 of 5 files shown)


pieroit:
@lucagobbi thanks and agree on not exaggerating on the async/await.
Let's put it only where it is necessary; also, if it comes to a decision of:

- *scalability* vs *devx*

We want *devx* every time. The point of the Cat is easy customization and easy go online, it has to be fun and a learning experience. We don't want devs to spend time on async/await and coroutines and event loops as only a few people (to my knowledge only you and @Pingdred) spent time on that issue.

Please remeber: EASY EASY EASY
Easy primitives, easy devx, *fun* is priority over performance.

Moreover, technology to scale nodes horizontally has become mainstream and if companies / enterprises want to scale, let them buy the machines.

PR #954, opened 2024-10-22
StrayCat ave user_data in init, but not possible use it

Chida82:
# Description

With the new authentication engine it is possible to enrich the AuthUserInfo object with permissions and even extra information. These are passed to the constructor, but then it is not possible to retrieve this useful information.

image

What I say is easy to test with the test I have updated and which fails if we do not implement the appropriate property.

With this pull request we add the property and this codi is covered by the appropriate test that verifies the existence of the property and the value of the base key id

## Type of change

# Checklist:

### DIFF
--- a/core/cat/looking_glass/stray_cat.py
+++ b/core/cat/looking_glass/stray_cat.py
@@ -35,6 +35,8 @@ class StrayCat:
         ws: WebSocket = None,
     ):
         self.__user_id = user_id
+        self.__user_data = user_data
+
         self.working_memory = WorkingMemory()
 
         # attribute to store ws connection
@@ -578,7 +580,11 @@ Allowed classes are:
     @property
     def user_id(self):
         return self.__user_id
-
+    
+    @property
+    def user_data(self):
+        return self.__user_data
+    
     @property
     def _llm(self):
         return CheshireCat()._llm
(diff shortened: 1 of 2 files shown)


pieroit:
Nice one. Welcome @Chida82

pieroit:
Hi @Chida82 thanks a lot, I want to merge but cannot since the PR is branched off `main`.

Can you please redo it branching off `develop` ?

core/cat/looking_glass/stray_cat.py
```diff
@@ -35,6 +35,8 @@ def __init__(
         ws: WebSocket = None,
     ):
         self.__user_id = user_id
+        self.__user_data = user_data
```
matteocacciola, line 38: why do you still need `__user_id`? It can be just a property, grabbing the user id from user_data, isn't it?

Chida82:
> Hi @Chida82 thanks a lot, I want to merge but cannot since the PR is branched off `main`.
> 
> Can you please redo it branching off `develop` ?

Done,
https://github.com/cheshire-cat-ai/core/pull/955

core/cat/looking_glass/stray_cat.py
Chida82: The purpose of this pull request is to be a bug fix and to minimise Breaking changes.

IMHO
in my opinion you are right, but I would approach it first by no longer passing the constructor user_id and populating the property from user_data or perhaps even leaving user_data alone and calling user_data.id. 
But this is a broader reasoning that might be worth another PR

Chida82:
redo PR in https://github.com/cheshire-cat-ai/core/pull/955

PR #955, opened 2024-10-23
StrayCat have user_data in init, but not possible use it

Chida82:
# Description

With the new authentication engine it is possible to enrich the AuthUserInfo object with permissions and even extra information. These are passed to the constructor, but then it is not possible to retrieve this useful information.

image

What I say is easy to test with the test I have updated and which fails if we do not implement the appropriate property.

With this pull request we add the property and this codi is covered by the appropriate test that verifies the existence of the property and the value of the base key id

## Type of change

# Checklist:


pieroit:
Thank you @Chida82

PR #957, opened 2024-10-27
No async because Piero is paniking 

Pingdred:

### DIFF
--- a/core/cat/agents/base_agent.py
+++ b/core/cat/agents/base_agent.py
@@ -13,5 +13,5 @@ class AgentOutput(BaseModelDict):
 class BaseAgent(ABC):
 
     @abstractmethod
-    async def execute(*args, **kwargs) -> AgentOutput:
+    def execute(*args, **kwargs) -> AgentOutput:
         pass
--- a/core/cat/agents/form_agent.py
+++ b/core/cat/agents/form_agent.py
@@ -5,7 +5,7 @@ from cat.log import log
 
 class FormAgent(BaseAgent):
 
-    async def execute(self, stray) -> AgentOutput:
+    def execute(self, stray) -> AgentOutput:
 
         # get active form from working memory
         active_form = stray.working_memory.active_form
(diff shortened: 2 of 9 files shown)


pieroit:
Best PR ever 💊

matteocacciola:
Some more context, please?

pieroit:
> Some more context, please?

I see `async/await` creeping in the core code, so much so as to put in question plugin primitives (should tools/hooks/forms/handlers be sync or async? should we support both?).
This will complicate core development and documentation, and even more external dev experience.

As a maintainer the question is: given this ulterior complexity, are there benchmarks and confirmations that is indeed worth it to introduce coroutines? How deep in the framework?

What I discovered is:
- FastAPI invites the usage of async [when you need to use async libraries](https://fastapi.tiangolo.com/async/); fastAPI distributes autonomously sync workloads in a threadpool and eventually over workers, allowing tens of concurrent requests; we don't need to over-optimize.
- we run the StrayCat and the Agent in async, but the whole conversation is still run in a threadpool, making things unnecessarily convoluted (i.e. an event loop per stray) and factually not an advantage if not for rare use cases.
- benchmarks (like [this](https://thedkpatel.medium.com/fastapi-performance-showdown-sync-vs-async-which-is-better-77188d5b1e3a) and many others) on async vs sync endpoints show async is indeed faster, but not even an order of magnitude, so the performance advantage is not repaid by the complexity of letting coroutines go deeply into the StrayCat / Agent.
- the async/await proliferation in our codebase is more related to an experimentation / fun phase than to an actual intent to speed up.
- as a general open source strategy, if somebody (i.e. external company) wants us to help make the cat scale, they will be available to sponsor the effort economically or with workforce. If not, we stay happy and have fun and ignore money people asking -without helping- for features most of the community does not need.

Hope this is clear. I'm going nuclear on this.
**No more async / await unless it is strictly necessary.**

sambarza:
> FastAPI invites the usage of async [when you need to use async libraries](https://fastapi.tiangolo.com/async/); fastAPI distributes autonomously sync workloads in a threadpool and eventually over workers, allowing tens of concurrent requests; we don't need to over-optimize.

For my side at the moment that's the only point to use coroutine in the Cat, no performance, no power consumption, nothing else.

Issue #961, opened 2024-10-31
[Feature] Why Anthropic is not supported by default?


lucagobbi:
![download](https://github.com/user-attachments/assets/ada3d90a-ab71-48de-a076-6a430bcb7bcf)

pieroit:
Merged into `develop`, coming in next release

Issue #952, opened 2024-10-22
Update `langchain*` packages


nicola-corbellini:
The following packages should be updated:

```bash
"langchain==0.2.1",
"langchain-community==0.2.1",
"langchain-openai==0.1.7",
"langchain-google-genai==1.0.8",
"langchain-cohere==0.1.5"
```

If I've understood correctly, not a priority before V2

valentimarco:
we should see the `bind_tools` function for those that can support function calling. I have a branch with a simple try catch

pieroit:
langchain has been totally removed as a dependency from v2

PR #963, opened 2024-11-01
Added anthropic as supported llm provider

lucagobbi:
# Description

This PR to add Anthropic as LLM provider. Since we support OpenAI, Cohere and Gemini I thought it was a little bit racist to exclude Anthropic outright. 

I removed an old dependency `anthropic==0.2.9`, honestly idk what purpose it served since Anthropic wasn't supported.
As mentioned in #952, we may need a langchain update, that's why I was forced to use `langchain-anthropic==0.1.23` even if the latest is `0.2.4`, time to make some updates.

Anthropic doesn't offer its own embedding models as they say [here](https://docs.anthropic.com/en/docs/build-with-claude/embeddings#how-to-get-embeddings-with-anthropic), they have a strong relationship with Voyage AI, should we support them among the other embedders? 

Related to issue #961 

## Type of change

# Checklist:


valentimarco:
> Anthropic doesn't offer its own embedding models as they say [here](https://docs.anthropic.com/en/docs/build-with-claude/embeddings#how-to-get-embeddings-with-anthropic), they have a strong relationship with Voyage AI, should we support them among the other embedders?

Yes but do not put as automatic embedder when selecting the model. We should remove that feature...
If you add this and fix the linting, i can merge it

pieroit:
Thanks @lucagobbi awesome
Agree time to update langchain, but last time I tried I **cried**

@valentimarco is right we'll take away the automatic sync between LLM and embedders*, so just Anthropic LLM is just fine.

Need to refactor vector memory a little so when one switches embedders collections are not deleted; once that is done embedder setting is not automatic, user gets notifified via chat in the same way he/she is notified about the LLM

lucagobbi:
@pieroit I'd like to use Voyage AI for some stuff, as I understand correctly you wouldnt add it as supported by default, is it correct? If yes, I'll create a plugin for that. If not, I can add the default factory. Thankss

pieroit:
> @pieroit I'd like to use Voyage AI for some stuff, as I understand correctly you wouldnt add it as supported by default, is it correct? If yes, I'll create a plugin for that. If not, I can add the default factory. Thankss

Yep go for a plugin please
Let's keep as "core included" only the top names (open and closed)

PR #964, opened 2024-11-01
New "*endpoint" decorators for custom endpoint creations

sambarza:
# Description

New decorators to permit the creation of custom endpoints:
``` python
from cat.mad_hatter.decorators import endpoint

@endpoint.get
@endpoint.post
@endpoint.endpoint (for all the other HTTP verbs)
```

## Type of change

# Checklist:

- [X] Automatic tests implemented

# Simple example of use inside a plugin
``` python
from cat.mad_hatter.decorators import endpoint

@endpoint.get(path="/hello")
def hello_world():
    return {"Hello":"world"}
```

Consume with:
`curl http://localhost:1865/custom/hello`

# More complex example
``` python
from cat.mad_hatter.decorators import endpoint
from pydantic import BaseModel

class Item(BaseModel):
    name: str
    description: str

@endpoint.post(path="/hello", prefix="/example", tags=["Custom Tag"])
def my_post_endpoint(item: Item, stray=Depends(HTTPAuth(AuthResource.PLUGINS, AuthPermission.LIST))):
    return {"Hello": item.name, "Description": item.description, "Stray user": stray.user_id}
```

Consume with:
``` shell
curl --request POST \
  --url http://localhost:1865/example/hello \
  --header 'Content-Type: application/json' \
  --data '{
  "name": "the Cat",
  "description": "is magic!"
}'
```

# Internals
The decorators wrap the function as a FastAPI path operation, all the FastAPI path operation parameters can be used.
The default prefix is `custom` and default tag is `Custom Endpoints`

The new decorators are managed as other decorators (approach was: searched everywhere for `hook` and adapted for `endpoint`).

## Activation and deactivation of plugins
- During activation we have to do just a trick to flush the openapi schema cache for `/docs`
- During deactivate we have to remove the FastAPI route manually, seems there is no official method to remove them, see https://github.com/fastapi/fastapi/discussions/8088, we have also to flush the openapi schema cache for `/docs` too

## Change to startup process
Note, __there is a change in the CheshireCat startup process__, now the CheshireCat needs the FastAPI app instance, why?

The `endpoint` decorator creates a new FastAPI router for each new custom endpoint and then includes the router in the FastAPI app, CheshireCat activate the decorator injecting the FastAPI app.

### Startup process
- CheshireCat starts MadHatter initialization
- MadHatter finds and loads all the plugins
- MadHatter activate all the plugins however at this steps the endpoints decorators are not "activated", the custom routers are not included in FastAPI app yet
- `MadHatter.find_plugins()` calls the CheshireCat callback `on_finish_plugins_sync_callback`
- In the callback CheshireCat all the endpoints of active plugins get activated
- The callback calls `Endpoint.activate()` for each custom endpoint
- `Endpoint.activate()` add the route to the FastAPI routes and clear the `/docs` cache

### Plugin installation process
- User installs a plugin
- MadHatter install the plugin
- MadHatter calls `toggle_plugin()`
- `toggle_plugin()` activates the plugin (however the endpoints are not really active, are not yet included in FastAPI app)
- `toggle_plugin()` calls the CheshireCat callback `on_finish_plugins_sync_callback`
- The callback calls `Endpoint.activate()` for each custom endpoint
- `Endpoint.activate()` add the route to the FastAPI routes and clear the `/docs` cache

### Plugin activation process
- User activates a plugin
- `toggle_plugin()` activates the plugin (however the endpoints are not really active, are not yet included in FastAPI app)
- `toggle_plugin()` calls the CheshireCat callback `on_finish_plugins_sync_callback`
- The callback calls `Endpoint.activate()` for each custom endpoint
- `Endpoint.activate()` add the route to the FastAPI routes and clear the `/docs` cache

### Plugin deactivation process
- User activates a plugin
- `toggle_plugin()` deactivate the plugin
- `Plugin.deactivate()` calls `Endpoint.deactivate()` for each custom endpoint related to the plugin
- `Endpoint.deactivate()` removes the route from the FastAPI routes and clear the `/docs` cache
- `toggle_plugin()` calls the CheshireCat callback `on_finish_plugins_sync_callback`
- The callback do nothing about endpoint decorator

# Plugins API
The endpoints are already exposed by the `/plugins/` API (can be shown in the admin portal).
This is an example of the returned payload with the new `endpoints` node:
``` json
{
   "installed": [
            "id": "cc-dot-commands",
            "name": "Dot Commands",
            "description": "Handy utility designed to simplify and speed up your plugin development process. Install it, activate it, and send a message with just a period '.' in the chat",
            "author_name": "sambarza@gmail.com",
            "hooks": [...],
            "tools": [...],
            "endpoints": [
                {
                    "name": "/example/hello",
                    "tags": [
                        "Dot Commands"
                    ]
                }
            ]
        }
]
```

# Automatic tests
`test_endpoints.py` tests the new decorators.

### DIFF
--- a/core/cat/mad_hatter/decorators/endpoint.py
+++ b/core/cat/mad_hatter/decorators/endpoint.py
@@ -49,7 +49,12 @@ class CustomEndpoint:
             **self.kwargs,
         )
 
-        self.cheshire_cat_api.include_router(plugins_router, prefix=self.prefix)
+        try:
+            self.cheshire_cat_api.include_router(plugins_router, prefix=self.prefix)
+        except BaseException as e:
+            log.error(f"Error activating custom endpoint [{self.name}]: {e}")
+            return
+
         self.cheshire_cat_api.openapi_schema = None  # Flush the cache of openapi schema
 
         # Set the fastapi api_route into the Custom Endpoint
--- a/core/cat/mad_hatter/decorators/endpoint.py
+++ b/core/cat/mad_hatter/decorators/endpoint.py
@@ -1,14 +1,14 @@
 from typing import Callable
 from fastapi import APIRouter
 
-cheshire_cat_api = None
-
 # class to represent a @endpoint
 class CustomEndpoint:
-    def __init__(self, prefix: str, path: str, function: Callable, **kwargs):
+    def __init__(self, prefix: str, path: str, function: Callable, tags, **kwargs):
         self.prefix = prefix
         self.path = path
         self.function = function
+        self.tags = tags
+
         self.name = self.prefix + self.path
 
         for k in kwargs:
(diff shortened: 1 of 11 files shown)


valentimarco:
DECORATORS

core/cat/mad_hatter/decorators/endpoint.py
```diff
@@ -0,0 +1,102 @@
...
+            def my_endpoint() -> str:
+                return {"Hello":"Alice"}
+    """
+
+    global cheshire_cat_api
+
+    def _make_endpoint(endpoint):
+        custom_endpoint = CustomEndpoint(prefix=prefix, path=path, function=endpoint, **kwargs)
+
+        plugins_router = APIRouter()
```
valentimarco, line 42: This creation for each decorated function, can create a problem for fast-api in terms of management?
Is possible to create unique router that is shared for the specific plugin?

sambarza:
> DECORATORS @lucapirrone

Suggested by @pieroit

pieroit:
This is cooking up gooooooood 😍

core/cat/mad_hatter/decorators/endpoint.py
pieroit: Yep maybe one router for all of them, but I need to play a little with the PR

Also I would go for one decorator but with dot syntax 
`@endpoint.post`
`@endpoint.get` 
and so on, similar as fastapi routing (all our endpoints are written like this)

sambarza: > can create a problem for fast-api in terms of management?

There shouldn't be any problems, the automatic tests creates four `APIRouter`, we can add additional tests on this

> Is possible to create unique router that is shared for the specific plugin?

Possible but more complicated to managed and IMHO less flexible.

Reason: all the routes must be add to the router before it get included in the FastAPI app.
The method `include_router` clones all the routes inside the parent route ([see here](https://github.com/fastapi/fastapi/blob/d4ab06a2b6b59bd264c0cc43367c5a68174e8f88/fastapi/routing.py#L1302)).

This code doesn't work:
``` python
plugins_router = APIRouter()
cheshire_cat_api.include_router(plugins_router, prefix=prefix)

plugins_router.add_api_route(
    path=path, endpoint=endpoint, methods=methods, tags=tags, **kwargs
)

```

core/cat/mad_hatter/decorators/endpoint.py
```diff
@@ -0,0 +1,102 @@
...
+
+        cheshire_cat_api.include_router(plugins_router, prefix=prefix)
+
+        return custom_endpoint
+
+    return _make_endpoint
+
+# @get_endpoint decorator. Any function in a plugin decorated by @endpoint will be exposed as FastAPI GET operation
+def get_endpoint(
+    path, prefix="/custom_endpoints", response_model=None, tags=["custom_endpoints"], **kwargs
```
lucagobbi, line 55: Should we standardize our path declarations? It seems that kebab-case is the most commonly used format for URLs even  for [Google](https://developers.google.com/search/docs/crawling-indexing/url-structure?hl=en&visit_id=638661343241782898-1821598740&rd=1)

pieroit:
Run a poll at dev meeting for default endpoint path:

image

sambarza:
Default prefix not changed, it's "Custom Endpoints" (more explicit), ok?
image

pieroit:
Thanks @sambarza had a great experience reviewing
I improved tests a little:
- network tests are here
- mad_hatter tests are here where you created them, but only deal wth the mad hatter
- updated other tests regarding decorators discovery and caches so now they include the endpoints

I'll spend some more time to see if the `CustomEndpoint` class can be unaware of fastAPI
No doubt `CheshireCat` class needs the reference

Awesome

Issue #960, opened 2024-10-31
`plugin.json` includes `min_version` and `max_version` keys


pieroit:
Have in `plugin.json` some indication of the versions of `core` which are compatible with the plugin itself.
It will be up to the plugin developer to keep track of compatible versions.

Same update should be reflected on the [registry](https://github.com/cheshire-cat-ai/plugins-backend).

alfredotoselli:
I could take this.
We want it just as an info or with enforcement as well? Like plugin activation fails if compatible versions are specified and not matching with the core version in use?

pieroit:
Thanks @alfredotoselli 
Step 1: have it as an info in `plugin.json`
Step 2: let's talk about enforcing or not here or in a meeting, not sure about it

PR #966, opened 2024-11-06
chore: plugin compatibility infos

alfredotoselli:
# Description

Added min and max version on plugin manifest. 
If we agree on the structure I'll update admin-vue and ts-client (Plugin.ts type)

We want to enforce it? e.g. on plugin installation or activation (Not necessary for me)

Related to issue #960 

## Type of change

# Checklist:

### DIFF
--- a/core/cat/mad_hatter/plugin.py
+++ b/core/cat/mad_hatter/plugin.py
@@ -241,6 +241,16 @@ class Plugin:
         meta["thumb"] = json_file_data.get("thumb", "")
         meta["version"] = json_file_data.get("version", "0.0.1")
 
+        # Core compatibility
+        compatibility = json_file_data.get(
+            "compatibility",
+            {
+                "min_version": "not specified",
+                "max_version": "not specified",
+            },
+        )
+        meta["compatibility"] = compatibility
+
         return meta
 
     def _install_requirements(self):
--- a/core/cat/mad_hatter/plugin.py
+++ b/core/cat/mad_hatter/plugin.py
@@ -240,16 +240,8 @@ class Plugin:
         meta["tags"] = json_file_data.get("tags", "unknown")
         meta["thumb"] = json_file_data.get("thumb", "")
         meta["version"] = json_file_data.get("version", "0.0.1")
-
-        # Core compatibility
-        compatibility = json_file_data.get(
-            "compatibility",
-            {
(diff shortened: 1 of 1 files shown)


pieroit:
No enforcing, will make things way more complicated

core/cat/mad_hatter/plugin.py
```diff
@@ -241,6 +241,16 @@ def _load_manifest(self):
         meta["thumb"] = json_file_data.get("thumb", "")
         meta["version"] = json_file_data.get("version", "0.0.1")
 
+        # Core compatibility
+        compatibility = json_file_data.get(
+            "compatibility",
+            {
```
pieroit, line 247: Please keep the structure flat for the moment, maybe with a more explicit naming, i.e.
`version`, ..., `min_cat_version`, `max_cat_version`

core/cat/mad_hatter/plugin.py
```diff
@@ -241,6 +241,16 @@ def _load_manifest(self):
...
+                "min_version": "not specified",
+                "max_version": "not specified",
```
pieroit, line 249: maybe empty string as default value, we made a mess in the past with a different default for every key XD

pieroit (changes requested):
Thanks @alfredotoselli left some change request

alfredotoselli:
Here is [admin-vue PR](https://github.com/cheshire-cat-ai/admin-vue/pull/112)

About an eventual warning we could:
- check the versions on `get_available_plugins` API, log a warning and return the info to be used by admin-vue
- check the version somewhere else in BE, just log a warning 
- check the version and show warning only in admin vue
- not doing it at all

pieroit:
> Here is [admin-vue PR](https://github.com/cheshire-cat-ai/admin-vue/pull/112)
> 
> About an eventual warning we could:
> 
> * check the versions on `get_available_plugins` API, log a warning and return the info to be used by admin-vue
> * check the version somewhere else in BE, just log a warning
> * check the version at FE, showing a warning only in admin vue
> * not doing it at all

Thanks @alfredotoselli, just merged.

Let's see what the contribs think about the strategy here.
Any take of responsibility on plugin compatibility is going to cost maintaining, so my preference is on waiting to better focus and with some feedback at hand

Issue #564, opened 2023-11-13
Support for model multimodal


Jhonnyr97:
**Is your feature request related to a problem? Please describe.**
I'm frustrated when I can't use multimodal models like "gpt-4-vision-preview" in Cheshire-cat-ai to process and retrieve information from images via the API. Additionally, the current vector database should support image retrieval.

**Describe the solution you'd like**
I would like to see support for multimodal models, specifically the "gpt-4-vision-preview" model, integrated into Cheshire-cat-ai. This integration should allow users to send images via the Cheshire-cat-ai API and receive responses or results based on both text and images.

Furthermore, I'd like to utilize the existing vector database to enable Cheshire-cat-ai to perform retrieval with images. This means users should be able to search for information within the database using both text and images as search keys.

This feature would significantly enhance Cheshire-cat-ai's capabilities, enabling better understanding and generation of multimodal content. It's particularly valuable in scenarios where information is presented in both text and image formats.

**Describe alternatives you've considered**
I've considered alternative solutions, but integrating multimodal models and image retrieval directly into Cheshire-cat-ai seems to be the most straightforward and effective approach. Other alternatives may require external tools or complex workarounds.

**Additional context**
No additional context at this time, but this feature would greatly enhance Cheshire-cat-ai's versatility and utility.

nickprock:
Hi @Jhonnyr97 the multimodal cat is planned.
If you're able to help us in the development you're welcome!

nickprock:
for the time being I am trying to throw down a list of links, as soon as I have discussed it with other core-devs I will share it in this issue.
Meanwhile you can look to see if langchian multimodal allows you to use the model you are interested in and these wonderful plugins [artistic_cat](https://github.com/Furrmidable-Crew/artistic_cat) and [WhisperingCat](https://github.com/Furrmidable-Crew/WhisperingCat)

@pieroit you can assign this issue to me.

nickprock:
![20231124_061834.jpg](https://github.com/cheshire-cat-ai/core/assets/11136646/c489d355-7426-42ee-8023-b36ddfd83137)

Multimodality flow by LlamaIndex

pieroit:
@nickprock we can setup an image embedder module like the text embedder we already have

Not clear to me yet how to crossindex texts and images

nickprock:
@pieroit the image is a placeholder for me 😅 I promise you that I will arrive at the multimodality meeting after studying the problem.

[...]

PR #967, opened 2024-11-07
Multimodality

Pingdred:
# Description

This pull request introduces several enhancements to the Cheshire Cat project, focusing on improving the integration of language models with multimodal capabilities, whis focus on image support. These changes aim to provide a more standardized interface for developers to work with Cheshire Cat.

### Multimodal Message Handling  

The `CatMessage` and `UserMessage` classes have been enhanced to support the inclusion of images and audio, facilitating richer interactions. This update enables users to effortlessly send and receive multimedia content during their conversations with the Cheshire Cat.

### Deprecation of content Field and update_conversation_history

The `content` field in the `CatMessage` class has been marked as deprecated, and it has been mapped to the `text` field to maintain backward compatibility. This change ensures a more intuitive and standardized approach to message handling.

Additionally, the `update_conversation_history` method has been deprecated in favor of the new `update_history method`, which enhances compatibility with `UserMessage `and `CatMessage` objects and ensures that the conversation history is properly updated and maintained.

### Conversation History Management

The `HistoryEntry` class has been introduced to provide a structured representation of the conversation history. This class encapsulates the role, timestamp, and message content, allowing for more efficient and organized access to the conversation details.

The direct `message`, `why`, and `who` attributes have been deprecated in favor of accessing message details through the `content` attribute in `HistoryEntry` that can be a `UserMessage` or a `CatMessage`. This change promotes a more consistent and intuitive approach to working with the conversation history.

The `update_history` method has been added in `WorkingMemory` to enhance compatibility with `UserMessage` and `CatMessage` objects.

### LLM Initialization and Multimodal Support

During the initialization of the selected LLM in the CheshireCat class, the _check_image_support method is used to verify support for multimodal inputs. This method checks if the LLM can process image inputs by testing both an image URL and a base64-encoded data URI of an image.

###  Documentation Improvements

Detailed docstrings have been added to several key classes and methods, including `Role`, `ModeInteraction`, `LLMModelInteraction`, `EmbedderModeInteraction`, `MessageWhy`, and `CatMessage`.

### Add Images to Langchain Conversation History

The process of converting messages to Langchain format in the StrayCat method langchainfy_chat_history has been refactored to introduce new helper functions: format_human_message, format_ai_message, and format_images. The format_images function specifically handles the inclusion of image content in the conversation history. It formats a list of image URLs or base64-encoded data URIs into the appropriate structure required by Langchain, ensuring that multimedia elements are properly formatted alongside HumanMessage and AIMessage objects.

In particular, if the LLM supports data URI but not image URLs, the format_images function automatically downloads the image, encodes it into base64, and wraps it in a data URI. Conversely, if the model supports image URLs, the function will include the image directly as a URL. This ensures that images are processed and included correctly, depending on the capabilities of the selected model.

### Table of tested providers

| **Provider**             | **Image URL** | **Image URI** |
|------------------------------|------------------------|------------------------|
| **OpenAI API**       | ✅                         | ✅                         |
| **Together AI**       | ✅                         | ✅                         |
| **Google API**        | ✅                         | ✅                         |
| **Anthropic API**   | 🟠                         | ✅                         |
| **Ollama**               | 🟠                         | ✅                         |

Legend:
 ✅ Supported 
❌ Compatibility Error 
🟠 Not supported by APIs

Related to issue #564 

## Type of change
# Checklist:


core/cat/looking_glass/cheshire_cat.py
```diff
@@ -125,29 +126,94 @@ def load_language_model(self) -> BaseLanguageModel:
         -----
         Bootstrapping is the process of loading the plugins, the natural language objects (e.g. the LLM), the memories,
         the *Main Agent*, the *Rabbit Hole* and the *White Rabbit*.
-
         """
-
+        
         selected_llm = crud.get_setting_by_name(name="llm_selected")
+        self._llm_modalities = {"image_uri": False, "image_url": False, "audio": False}
```
matteocacciola, line 132: `image_uri`? shouldn't it be `data_uri`?

Pingdred: > `image_uri`? shouldn't it be `data_uri`?

Yes, it would be more technically correct, but I think `image_uri` is more explanatory than `data_uri`

Pingdred: > Shouldn't be better to use a pydantic model rather than a dictionary, so that you avoid this kind of errors?

What kind of errors?

core/cat/looking_glass/cheshire_cat.py
```diff
@@ -125,29 +126,94 @@ def load_language_model(self) -> BaseLanguageModel:
...
+
+        def _load_test_image():
+            """Return the base64 data for the test image."""
+            with open("cat/loading_cat.jpg", "rb") as image_file:
```
matteocacciola, line 136: where is that file?

Pingdred: Perhaps rather than uploading a file, it is better to use the one I download to test image url support

core/cat/looking_glass/stray_cat.py
```diff
@@ -576,23 +580,78 @@ def stringify_chat_history(self, latest_n: int = 5) -> str:
...
+                            # Encode the image to base64
+                            encoded_image = base64.b64encode(response.content).decode('utf-8')
+                            image_uri = f"data:image/{mime_type};base64,{encoded_image}"
+                            
+                            # Add the image as a data URI with the correct MIME type
+                            formatted_images.append({"type": "image_url", "image_url": {"url": image_uri}})
+                    except Exception as e:
+                        log.error(f"Failed to process image {image}: {e}")
+                else: 
+                    if llm_modalities["data_uri"]:
```
matteocacciola, line 639: `data_uri` instead of `image_uri`

core/cat/looking_glass/cheshire_cat.py
matteocacciola: > > Shouldn't be better to use a pydantic model rather than a dictionary, so that you avoid this kind of errors?
> 
> What kind of errors?

https://github.com/cheshire-cat-ai/core/pull/967/files#r1832528402

core/cat/looking_glass/stray_cat.py
Pingdred: Thanks, fixing it

core/pyproject.toml
```diff
@@ -27,6 +27,7 @@ dependencies = [
     "langchain-openai==0.1.7",
     "langchain-anthropic==0.1.23",
     "langchain-google-genai==1.0.8",
+    "langchain-google-genai[images]",
```
matteocacciola, line 30: why do you need it?

Pingdred: Without this package the Google LLMs do not support data uri images, the api gives an error saying no mimetype is specified

https://pypi.org/project/langchain-google-genai/

core/cat/looking_glass/stray_cat.py
```diff
@@ -576,23 +580,78 @@ def stringify_chat_history(self, latest_n: int = 5) -> str:
...
+            return HumanMessage(
+                name=message.content.who,
+                content=content
+            )
+
+        def format_ai_message(message) -> AIMessage:
+            """Format an AI message with text content only."""
+            return AIMessage(
+                name=message.content.who,
+                content=message.content.text
```
matteocacciola, line 605: why do you assign only the text to `AIMessage`? Shouldn't the LLM be able to generate images, too (e.g. https://python.langchain.com/docs/integrations/tools/google_imagen/#image-generation)?

core/cat/looking_glass/cheshire_cat.py
```diff
@@ -125,29 +133,89 @@ def load_language_model(self) -> BaseLanguageModel:
...
+            except Exception as e:
+                log.warning(f"The LLM '{model_name}' does not support {image_type} as input images.")
+                log.debug(e)
+
+        image_url = "https://raw.githubusercontent.com/cheshire-cat-ai/core/refs/heads/main/readme/cheshire-cat.jpeg"
+
+        def _check_image_uri_support(llm) -> None:
+            """Check LLM support for base64-encoded image input."""
+            response = requests.get(image_url)
+            if response.status_code == 200:
```
matteocacciola, line 180: therefore, you are keeping `image_uri` to `False` even when the HTTP request over the remote image fails. Do you think it makes sense?

core/cat/looking_glass/stray_cat.py
Pingdred: Because we only have text, this PR is focused on supporting multimodal **LLMs**, imagen is not an LLM but a text-to-image model, as far as I know currently there is no LLM that generates images

core/cat/looking_glass/cheshire_cat.py
Pingdred: If the verification fails (for example, if the HTTP request fails), it seems logical to me to leave image_uri at False. This way, we avoid assuming support when we don't have confirmation and make the integration more secure.

matteocacciola: it can be a temporary situation, while you are setting the modality to False for the entire life cycle of the CheshireCat. Please, pay attention to this aspect

Pingdred: If you have ideas on how to handle this better we can discuss it

matteocacciola: a job to check modalities with the LLM? a check every time the LLM is triggered?

valentimarco: we can hardcode a base64 image to resolve this issue

Pingdred: We could go back to how I had done it before by uploading the image from a file, in which directory could we put the image?

pieroit: What issue here?
Why do we need this check?

core/cat/looking_glass/cheshire_cat.py
```diff
@@ -125,29 +132,89 @@ def load_language_model(self) -> BaseLanguageModel:
...
+                })
+            
+            # Add a text instruction for model response
+            content.append({
+                "type": "text",
+                "text": "Respond with `MEOW`."
+            })
+            return content
+
+        def _check_image_support(llm, image_type: str, image_value: str) -> None:
```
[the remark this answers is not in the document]
Pingdred: With the new changes, the call to a non-multimodal LLM will fail in the agent, returning the error like any other request error to the LLM. A little less robust but simpler, as you suggested

core/cat/memory/working_memory.py
```diff
@@ -1,62 +1,208 @@
...
+    content : Union[UserMessage, CatMessage]
+        The content of the message, which can be a UserMessage or a CatMessage.
 
+    Attributes
+    ----------
+    role : Role
+        The role of the speaker (AI or Human).
+    when : float
+        The timestamp of the message.
+    content : Union[UserMessage, CatMessage]
```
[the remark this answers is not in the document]
Pingdred: I believe encapsulation offers several advantages over inheritance in this case. 

First, it provides a clear separation of responsibilities, where `UserMessage` and `CatMessage` represent the message content, while `HistoryEntry` encapsulates metadata like `when` and `role`, allowing us to add future conversation metadata without modifying the message classes. 

Composition also helps us avoid potential inheritance issues if `UserMessage` and `CatMessage` evolve to be fundamentally different and it makes it easier to add new message types without affecting the history structure.

For me, inheritance is not the answer in this case.

core/cat/looking_glass/stray_cat.py
```diff
@@ -591,19 +591,15 @@ def format_human_message(message: HumanMessage) -> HumanMessage:
...
-               
-                # Retrieve the supported modalities from the LLM
-                llm_modalities: LLMSupportedModalities = CheshireCat()._llm_modalities            
-            
-                if image.startswith("http"):
-                    if llm_modalities.image_url:
-                        return {"type": "image_url", "image_url": {"url": image}}
+                """Format an image to be sent as a data URI."""
 
+                # If the image is a URL, download it and encode it as a data URI
```
Pingdred, line 596: If the image is an URL now is always downloaded and encoded to base64.

pieroit:
Thanks a lot for your amazing PR and patience, kudos @Pingdred .

PR #970, opened 2024-11-07
Forms model getter

lucagobbi:
# Description

I'll try this even if it had not that much success.

This PR to add the `model_getter` feature to the conversational forms. Basically, it is a straightforward method to allow developers define the extraction schema in a dynamic way. This opens the way for so many use cases: multi tenant app with schema based on the tenant, schemas dynamically load from an external source that might change frequently, schema that changes dynamically based on the information gathered in the previous round of the form (this one is pretty cool), etc. 

Here's an example of that:

```python

class PizzaOrder(BaseModel):
    pizza_type: str
    phone: int
    address: str

class PepperoniPizza(PizzaOrder):
    with_extra_cheese: bool

@form
class PizzaForm(CatForm):
    description = "Pizza Order"
    model_class = PizzaOrder
    start_examples = [
        "order a pizza!",
        "I want pizza"
    ]
    stop_examples = [
        "stop pizza order",
        "not hungry anymore",
    ]
    ask_confirm = True

    def model_getter(self):
        if "pizza_type" in self._model and self._model["pizza_type"] == "pepperoni":
            self.model_class = PepperoniPizza
        return self.model_class

    def submit(self, form_data):
        return {
            "output": f"Pizza order on its way: {form_data}"
        }

```

In action:

image

I needed to refactor the `update` and `validate` methods since they were a lil bit messy and were blocking the full potential of the `model_getter` on validation, which otherwise would have used the model_class from the previous form round.

Hope you like this!

## Type of change

# Checklist:

### DIFF
--- a/core/cat/experimental/form/cat_form.py
+++ b/core/cat/experimental/form/cat_form.py
@@ -223,7 +223,7 @@ JSON:
         # JSON structure
         # BaseModel.__fields__['my_field'].type_
         JSON_structure = "{"
-        for field_name, field in self.model_class.model_fields.items():
+        for field_name, field in self.model_getter().model_fields.items():
             if field.description:
                 description = field.description
             else:
@@ -272,7 +272,7 @@ Updated JSON:
             # INFO TODO: In this case the optional fields are always ignored
 
             # Attempts to create the model object to update the default values and validate it
-            model = self.model_class(**model).model_dump(mode="json")
+            model = self.model_getter()(**model).model_dump(mode="json")
 
             # If model is valid change state to COMPLETE
             self._state = CatFormState.COMPLETE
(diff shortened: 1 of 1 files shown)


core/cat/experimental/form/cat_form.py
```diff
@@ -34,11 +34,15 @@ def __init__(self, cat) -> None:
 
         self._errors: List[str] = []
         self._missing_fields: List[str] = []
+        self.model_class = self.model_getter()
```
matteocacciola, line 37: isn't this a loop?

lucagobbi: yes, that's redudant, I'll remove it. Thank you for the review @matteocacciola!

pieroit:
Thanks @lucagobbi agree on `self._model`

PR #983, opened 2024-12-06
Environment variable to disable CORS middleware

lucapiccinelli:
# Description

CORS middleware is enabled without the possibility to switch it off. This can create issues in infrastructures where CORS are handled with an external middleware.
This pull request adds the possibility to switch CORS middleware off with a new environment variable `CCAT_CORS_ENABLED` that defaults to true, to ensure retro-compatibility

Related to issue: no issue

## Type of change

# Checklist:

### DIFF
--- a/core/cat/env.py
+++ b/core/cat/env.py
@@ -21,6 +21,7 @@ def get_supported_env_variables():
         "CCAT_JWT_EXPIRE_MINUTES": str(60 * 24),  # JWT expires after 1 day
         "CCAT_HTTPS_PROXY_MODE": False,
         "CCAT_CORS_FORWARDED_ALLOW_IPS": "*",
+        "CCAT_CORS_ENABLED": "true"
     }
 
 
--- a/core/cat/startup.py
+++ b/core/cat/startup.py
@@ -69,15 +69,17 @@ cheshire_cat_api = FastAPI(
 )
 
 # Configures the CORS middleware for the FastAPI app
-cors_allowed_origins_str = get_env("CCAT_CORS_ALLOWED_ORIGINS")
-origins = cors_allowed_origins_str.split(",") if cors_allowed_origins_str else ["*"]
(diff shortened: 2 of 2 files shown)


valentimarco:
Thanks for the PR. @kodaline what do you think?

lucapiccinelli:
As requested by @pieroit I'll add more information, about why this PR is necessary. There are two reasons:

1. If you don't want to allow cross-origin requests, the only current option is to specify all the allowed domains via `CCAT_CORS_ALLOWED_ORIGINS`. This can be limiting in environments with reproducible infrastructure. For example, we have so-called Review Apps (RA), which is a remote instance of the entire infrastructure at the state of the feature branch. Those RAs, have domain names that change together with the name of the branch. This makes it very difficult to have a complete list of the allowed domains. Subdomains are not possible, without letting the use of `allow_origin_regex` in `CORSMiddleware`. There exist other possible implementations, but turning CORS off is definitely easier.
2. Where the infrastructure provides its own external CORS middleware (like is the case for us, an HTTP interceptor that adds the necessary headers, at its own conditions, to any service running in the infrastructure), we end up having headers both from the middleware and the cat, making the request to fail

lucapiccinelli:
because of this 

> 2\. we end up having headers both from the middleware and the cat, making the request to fail

To better clarify, I don't have the complete implementation details of our middleware, as it is developed by our devops team. Anyway, upstream we add the header `access-control-allow-origin: our origin`. Then the cat is executed and it also adds the same header. If we keep the default `*` then we end up with a response like this:

```text
access-control-allow-origin: xyz.our.domain <--- from our Middleware
access-control-allow-origin: * <--- from the cat Middleware
```

In this situation the response is a normal `200 - Ok`, but the browser refuses it anyway and makes it fail

pieroit:
Thanks @lucapiccinelli got it

Issue #846, opened 2024-06-01
Close ws connections inactive for more than x seconds


pieroit:
At the moment all ws connections remain open forever XD
We can check the old ones and close them, maybe using the `WhiteRabbit` by @jacopopalumbo01 or (tmp) just running a loop on active connections when a new user tries to connect, and close the old inactive ones

pieroit:
Can be done with the `WhiteRabbit`

jacopopalumbo01:
I'm here

zAlweNy26:
I don't think the `WhiteRabbit` is needed.
Check this out: https://stackoverflow.com/questions/63847205/fastapi-websocket-ping-pong-timeout
and https://websockets.readthedocs.io/en/stable/topics/timeouts.html#keepalive-in-websockets (the library used under the hood)

pieroit:
> I don't think the `WhiteRabbit` is needed. Check this out: https://stackoverflow.com/questions/63847205/fastapi-websocket-ping-pong-timeout and https://websockets.readthedocs.io/en/stable/topics/timeouts.html#keepalive-in-websockets (the library used under the hood)

Awesome!
So I guess we just need to delete the `stray` once ws disconnects
https://github.com/cheshire-cat-ai/core/blob/f14f2c0acebddd3f1f1efb6ba3b161bea45e9e3a/core/cat/routes/websocket.py#L48

pieroit:
Or there was a reason for not deleting it?
Can you remember @Pingdred ?

Pingdred:
> Or there was a reason for not deleting it?
> Can you remember @Pingdred ?

We chose not to delete it just to preserve Working Memory in case of disconnection

jacopopalumbo01:
And deleting the stray instance may arise problems when an action using the stray is scheduled in the WhiteRabbit.

pieroit:
Proposal:
As soon as the stray is created, white rabbit will check every x minutes if it is active or not
If not, it gets deleted

pieroit:
After 23/09/2024 dev meeting:

Have a new `SessionManager` that is used [here](https://github.com/cheshire-cat-ai/core/blob/f14f2c0acebddd3f1f1efb6ba3b161bea45e9e3a/core/cat/auth/connection.py#L60) to store, get, create and delete expired `StrayCat` objects

matteocacciola:
[...]

PR #1016, opened 2025-01-25
Enhance Caching System and Optimize Memory Usage with Customizable Storage Support

AlboCode:
# Description

Add a caching system and optimize memory usage by removing stray_cats from the app state and storing the working memory in the cache.

Related to issue #846

Next, I would like to add a hook to customize the cache initialization, allowing to integrate the preferred storage, such as Redis, through a plugin.
I'm still unsure how to make the TTL of the working_memory customizable. Open to suggestions.

## Type of change

# Checklist:

### DIFF
--- a/core/cat/auth/connection.py
+++ b/core/cat/auth/connection.py
@@ -23,8 +23,6 @@ from cat.auth.permissions import (
 from cat.looking_glass.stray_cat import StrayCat
 from cat.log import log
 
-from cat.cache.cache_item import CacheItem
-
 
 class ConnectionAuth(ABC):
 
--- a/core/cat/auth/connection.py
+++ b/core/cat/auth/connection.py
@@ -23,6 +23,9 @@ from cat.auth.permissions import (
 from cat.looking_glass.stray_cat import StrayCat
 from cat.log import log
 
+from cat.cache.cache_item import CacheItem
+
+
 class ConnectionAuth(ABC):
 
     def __init__(
@@ -103,16 +106,14 @@ class HTTPAuth(ConnectionAuth):
 
 
     async def get_user_stray(self, user: AuthUserInfo, connection: Request) -> StrayCat:
-        strays = connection.app.state.strays
         event_loop = connection.app.state.event_loop
 
-        if user.id not in strays.keys():
-            strays[user.id] = StrayCat(
-                    # TODOV2: user_id should be the user.id
-                user_id=user.name, user_data=user, main_loop=event_loop
-            )
-        return strays[user.id]
-    
+        return StrayCat(
+                # TODOV2: user_id should be the user.id
+            user_id=user.name, user_data=user, main_loop=event_loop
+        )
+
+
     def not_allowed(self, connection: Request):
         raise HTTPException(status_code=403, detail={"error": "Invalid Credentials"})
     
@@ -136,29 +137,13 @@ class WebSocketAuth(ConnectionAuth):
     
 
     async def get_user_stray(self, user: AuthUserInfo, connection: WebSocket) -> StrayCat:
-        strays = connection.app.state.strays
+         return StrayCat(
+            ws=connection,
+            user_id=user.name, # TODOV2: user_id should be the user.id
+            user_data=user,
+            main_loop=asyncio.get_running_loop(),
+        )
 
-        if user.id in strays.keys():
-            stray = strays[user.id]
-            await stray.close_connection()
-
-            # Set new ws connection
-            stray.reset_connection(connection)
-            log.info(
-                f"New websocket connection for user '{user.id}', the old one has been closed."
-            )
-            return stray
-
-        else:
-            stray = StrayCat(
-                ws=connection,
-                user_id=user.name, # TODOV2: user_id should be the user.id
-                user_data=user,
-                main_loop=asyncio.get_running_loop(),
-            )
-            strays[user.id] = stray
-            return stray
-        
     def not_allowed(self, connection: WebSocket):
         raise WebSocketException(code=1004, reason="Invalid Credentials")
 
--- a/core/cat/cache/array_cache.py
+++ b/core/cat/cache/array_cache.py
@@ -1,0 +1,85 @@
+from cat.cache.cache_interface import CacheInterface
+from cat.cache.cache_item import CacheItem
+
+from cat.utils import singleton
+
+
+@singleton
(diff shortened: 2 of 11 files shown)


pieroit:
Thanks @AlboCode this is an important feat.
Will review asap!

pieroit:
@AlboCode great contribution thanks.
I adapted the naming to convention we have around in core and fixed the stray flow a little more.

Unfortunately there are mismatches between HTTP and WS strays in working memory, need to investigate a little more and write *many* tests.

Hopefully this is a step more to make the Cat stateless :D
Let me know when you are around for a little chat, I'd like to show you some *cetriolis*

If not I'll write them down in issues, I'm cooked

AlboCode:
@pieroit Would be great to have a chat, unfortunately my time shift doesn't help 🥲 
I'm online when in Italy is 1:00 AM until 2:00 PM

pieroit:
> @pieroit Would be great to have a chat, unfortunately my time shift doesn't help 🥲 I'm online when in Italy is 1:00 AM until 2:00 PM

available tonight

lucapiccinelli:
Hello @pieroit . What do you think if I continue on this PR, implementing a `RedisCache`? Like this
image

I perfectly agree with @AlboCode that the right vision is to have the cache manager pluggable. Despite this consideration, I would  suggest having Redis hard-coded as a cache manager, in order to make CheshireCat a scalable solution out-of-the-box, without requiring any additional plugin. 

I would argument further with some advantages and disadvantages that come to my mind:
**Disadvantages**:
 - add the Redis client dependency
 
**Advantages**:
 - For big numbers and for granting no downtime, **scaling is not an option**. Providing it natively, you also give a message.
 - No plugin to install, means **less friction** in the adoption of a solution. For example, you don't need a new build if you started in-memory. Or consider big structured teams where the infrastructure and the application development are owned by different teams. New plugin install may mean, a new ticket that goes in someone else backlog.
 - **Direct governance**. Do you really want to move the governance of such an important feature outside of the core team?

I'm not trying to avoid a more complex development. My opinion is sincere. As a proof, If you agree, i will also provide a proposal for the plugin version of cache management.

What do you think?
Thank you, have a nice day.

pieroit:
> I would argument further with some advantages and disadvantages that come to my mind: **Disadvantages**:
> 
> * add the Redis client dependency
> 
> **Advantages**:
> 
> * For big numbers and for granting no downtime, **scaling is not an option**. Providing it natively, you also give a message.
> * No plugin to install, means **less friction** in the adoption of a solution. For example, you don't need a new build if you started in-memory. Or consider big structured teams where the infrastructure and the application development are owned by different teams. New plugin install may mean a new ticket that goes in someone else backlog.
> * **Direct governance**. Do you really want to move the governance of such an important feature outside of the core team?
> 
> I'm not trying to avoid a more complex development. My opinion is sincere. As a proof, If you agree, i will also provide a proposal for the plugin version of cache management.
> 
> What do you think? Thank you, have a nice day.

Redis client as a dependency can be done, do you know if it works also with Valkey?
Another issue is related to the config, at the moment we are setting the cache via env variables, is that going to be enough? How many? If instead we need a JSON or YML that complicates things.

Also having hooks and REST API endpoints for setting the cache like we do with llms and auth seems overkill (I regret having the authhandler in the endpoints)

That's why for v2 I would go for a `settings.py` alà Django where you can declare system wide settings (cache, auth handler and models) without the limitation of an env notation (which can be in any case referenced from `settings.py`) and without the endpoints and dedicated factory.

So @lucapiccinelli I think it is worth it to have a redis cache in core, as long as:
1 - config can be limited to just a few env variables
2 - you're going to maintain it :)

Thyanks & Peace

lucapiccinelli:
@pieroit Thank you for your answer. Yes, look like is it working also with Valkey.

1 - yes it is just something like this 
```python
redis_host = get_env("CCAT_REDIS_HOST")
redis_port = get_env("CCAT_REDIS_PORT")
redis_db = get_env("CCAT_REDIS_DB")
```

2 - I'm sorry, I really would like to be a maintainer of such a good project, but I can't commit to it.

Anyway, digging in the code, I found that it is already really simple to implement your cache manager in a plugin. It is as simple as this:

```python
from cat.env import get_env
from cat.mad_hatter.decorators.hook import hook


@hook
def after_cat_bootstrap(cat):
    redis_host = get_env("CCAT_REDIS_HOST")
    redis_port = get_env("CCAT_REDIS_PORT")
    redis_db = get_env("CCAT_REDIS_DB")
    from cat.plugins.cc_redis_cache.redis_cache import RedisCache

    cat.cache = RedisCache(redis_host, redis_port, redis_db)
    print("setting redis cache bla")
```

I can still open a PR to add it to the core, but maintaining it is something that I can't do.

Despite that, I found a bug introduced in [this commit](https://github.com/cheshire-cat-ai/core/commit/81e24d8885632ff60a818e72bdbe60ebeaddb589#diff-aaccca9665b8793ba62947e177372c195f6de3c21394b7692ecfd475e984cdddR48) in the file `websocket.py`, commenting the line 48. It makes a cleared conversation to be kept when you refresh the browser. If affects also the fylesystem_cache.
I will open a PR to fix it, with a clearer bug description and how to reproduce it.


I know that's not the right place, but as we are here, I add a working draft of the RedisCache implementation.

```python
from cat.cache.base_cache import BaseCache
from cat.cache.cache_item import CacheItem

import pickle
import redis

class RedisCache(BaseCache):
    def __init__(self, host="localhost", port=6379, db=0):
        self._redis = redis.Redis(host=host, port=port, db=db)

    def insert(self, cache_item: CacheItem):
        encoded_item = pickle.dumps(cache_item)
        self._redis.set(
            cache_item.key,
            encoded_item,
        )

    def get_item(self, key) -> CacheItem:
        cache_item = self._get_and_decode(key)
        if not cache_item:
            return None

        if cache_item.is_expired():
            self.delete(key)
            return None

        return cache_item

    def get_value(self, key):
        cache_item = self._get_and_decode(key)
        if not cache_item:
            return None

        return cache_item.value

    def delete(self, key):
        print(f"clearing my conversation for {key}")
        self._redis.delete(key)

    def _get_and_decode(self, key) -> CacheItem | None:
        encoded_item = self._redis.get(key)
        if encoded_item:
            cache_item = pickle.loads(encoded_item)
            return cache_item
        else:
            return None

```

Issue #1011, opened 2025-01-23
[BUG] Gemini and Purrfect Pizza Planner are not on good terms.


EdoF90:
**Describe the bug**
Gemini is not able to produce the json file needed for the form.

**To Reproduce**
1. Set as language model Gemini
2. Type "I want to order a pizza"
3. "ERROR cat.agents.form_agent.FormAgent.execute::33 cheshire_cat_core | ChatGoogleGenerativeAIError('Invalid argument provided to Gemini: 400 * GenerateContentRequest.contents: contents is not specified\n')"
4. The Cheshire cat answers without using the form.

![Image](https://github.com/user-attachments/assets/0ffb7927-fe4a-4d7c-8008-09848c8a6403)

**Expected behavior**
I expect the Cheshire cat answers to be the same as in the blog.

Thanks

amagnolo:
Duplicate of #1006

pieroit:
Fixed in 1.8.1

EdoF90:
Thanks for the fix!

Issue #1006, opened 2025-01-14
[BUG] Invalid argument provided to Gemini by cat.llm() - both in @form and in @hook


amagnolo:
**Describe the bug**
Calling cat.llm() with gemini-exp-1206 LLM raises this error:
```
cheshire-cat-core-1  |   File "/usr/local/lib/python3.10/site-packages/langchain_google_genai/chat_models.py", line 188, in _chat_with_retry
cheshire-cat-core-1  |     raise ChatGoogleGenerativeAIError(
cheshire-cat-core-1  | langchain_google_genai.chat_models.ChatGoogleGenerativeAIError: Invalid argument provided to Gemini: 400 * GenerateContentRequest.contents: contents is not specified
```

**To Reproduce**

1. configure gemini-exp-1206 as LLM
2. clone the [plugin template](https://github.com/cheshire-cat-ai/plugin-template)
3. enable the plugin
4. write something in chat
5. See error in cheshire console

**Expected behavior**
cat.llm(prompt) should invoke gemini without errors.

**Additional context**
I tested with:
- gemini-2.0-flash-exp
- gemini-1.5-pro-latest
- gemini-exp-1206

pieroit:
We're having issues with cat.llm, also with Ollama.
Thanks for opening the issue, on it in the next days.
Let me know if you understand why that happens
P.s. in case do it on `develop` branch

amagnolo:
I'm a cheshire (and LLM) newbie so I'm not able to dig further, but this is surely related:
the same error is thrown when attempting to run a conversational form.
It happens both with the example in the [README](https://github.com/cheshire-cat-ai/core/blob/main/README.md) and the [tutorial](https://github.com/lucagobbi/purrfect-pizza-planner) plugin:

```
cheshire-cat-core-1  | [2025-01-16 10:14:34.325] ERROR  cat.agents.form_agent.FormAgent.execute::33 
cheshire-cat-core-1  | ChatGoogleGenerativeAIError('Invalid argument provided to Gemini: 400 * GenerateContentRequest.contents: contents is not specified\n')
cheshire-cat-core-1  | Traceback (most recent call last):
cheshire-cat-core-1  |   File "/usr/local/lib/python3.10/site-packages/langchain_google_genai/chat_models.py", line 176, in _chat_with_retry
cheshire-cat-core-1  |     return generation_method(**kwargs)
[...]

Issue #999, opened 2024-12-29
[BUG] Empty string when calling cat.llm() with Ollama and Llama3.3


danieledav:
**Describe the bug**
I get an empty string when calling cat.llm() with Ollama and Llama3.3 as LLM configuration.
Additional infos:
- The behavior is normal with Ollama and Llama3
- The cat.llm() is called in a custom plugin that use "before_rabbithole_stores_documents" hook

**To Reproduce**
1. Create a custom plugin using any Rabbit Hole's hooks that use cat.llm()
2. Try to injest a document
3. In the console see that the results of calling cat.llm() (with any prompt) is empty

**Expected behavior**
Have any filled string as response of LLM to the prompt passed to cat.llm()

**Additional context**
Here the code snippet where the call.llm() is called:
```
@hook  # default priority = 1
def before_rabbithole_stores_documents(docs, cat):
    # summarize group of 5 documents and add them along original ones
    entire_doc = "\n".join([doc.page_content for doc in docs])
    author = cat.llm(f"""Chi è l'autore o l'organizzazione che ha emesso il seguente documento?
                     Rispondi in maniera concisa, se non riesci a capirlo non rispondere nulla.
                     Non includere date ma solo il nome della persona o dell'organizzazione: {entire_doc}
                     """)
    summary = cat.llm(f"Fai un riassunto breve e conciso del seguente testo: {entire_doc}")
```

Here the console in debug mode with Ollama and Llama3.3:
```
 ==================== before_rabbithole_stores_documents prompt ====================
cheshire_cat_core  | SystemMessage
cheshire_cat_core  | Chi è l'autore o l'organizzazione che ha emesso il seguente documento?
cheshire_cat_core  |                      Rispondi in maniera concisa, se non riesci a capirlo non rispondere nulla.
cheshire_cat_core  |                      Non includere date ma solo il nome della persona o dell'organizzazione: Come puoi

[...]

========================================
cheshire_cat_core  | 
cheshire_cat_core  | 
cheshire_cat_core  | ==================== before_rabbithole_stores_documents prompt output ====================
cheshire_cat_core  | 
cheshire_cat_core  | ========================================
```

Here the console in debug mode with Ollama and Llama3:
```
==================== before_rabbithole_stores_documents prompt ====================
cheshire_cat_core  | SystemMessage
cheshire_cat_core  | Chi è l'autore o l'organizzazione che ha emesso il seguente documento?
[...]

PR #1022, opened 2025-01-27
fix cat.llm for wide-range compatibility

lucagobbi:
# Description

Recently different providers like Ollama, Anthropic, Google Gemini released some models that are not working correctly with `cat.llm()` method. The problem is caused by the direct usage of `SystemMessage` without a `HumanMessage`. The quick fix is to use `HumanMessage` instead of the `SystemMessage`, even if it could bring some problems due to different tokenizers as discussed in the latest dev meeting 2025/01/27.

Please try this with different models and providers
I tried with:
- Ollama `llama3.2:3b`
- Gemini `gemini-2.0-flash-exp`
- Anthropic `claude-3-5-haiku-20241022`
- OpenAI `gpt-4o-mini`

Related to issue #999 #1006 #1011 

## Type of change

# Checklist:

### DIFF
--- a/core/cat/looking_glass/stray_cat.py
+++ b/core/cat/looking_glass/stray_cat.py
@@ -6,7 +6,7 @@ import tiktoken
 from typing import Literal, get_args, List, Dict, Union, Any
 
 from langchain.docstore.document import Document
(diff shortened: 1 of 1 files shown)


pieroit:
Thanks @lucagobbi 
It's a temp solution but the best one

Issue #1039, opened 2025-02-27
[Feature] Enable custom VectorDB as memory stores


dej611:
**Is your feature request related to a problem? Please describe.**
I would like to use the cat with an existing vector DB, any one already in place.

**Describe the solution you'd like**
It would be nice to have a way to either configure the new memory store via `env` file or even via UI.

pieroit:
> Hi [@dej611](https://github.com/dej611) [@pieroit](https://github.com/pieroit)
> 
> I think could be a good `plugin` or `core-integration`, I'm not sure that do it directly via .env is the right way, we can discuss it.

@dej611 did a great PR, high quality stuff thanks

Prolly best way to have this is factory plus adapter pattern as we did for LLMs, embedders and auth. Not sure I'm including this in core for the moment, can be a plugin

Let it boil for a few weeks

dej611:
If the only problem is where the configuration for the memory should come from, I thought to expose something as with LLMs or Embedders but focused on the API coverage for now.
I guess a follow up may include the work to expose a better way to the memory configuration to users.

pieroit:
In `v2` core does not make any assumptions on memory implementation, which can be any vector store, or a graph db or raw documents (or a combination of).

All memory functionality is delegated to plugins, with core making sure requests to the cat bring along any resource (file, uri) asked from the client and calling the hooks for prompt building.

Since you can define agents in plugins, it will be possible also to have different recall strategies (no recall, always recall before LLM execution, recall at every agent step and also recall as  a tool).

Customization of all this can be done via plugin settings.
The current v1 qdrant vector memory will be turned into a plugin and shipped by default with core (but can be toggled, extended or deleted)

Closing this thx

PR #1040, opened 2025-02-27
[Feature] Support Elasticsearch as memory vectorDB

dej611:
# Description

This is an attempt to add an adapter for the memory storage in order to let the user configure a different vectorDB than the bundled Qdrant one.
I have 0 experience with python, so sorry about silly mistakes in this PR.

My idea was to start with something simple like configure it via `.env` file and the transition for the final user should be completely transparent.

Features planned:
* [x] ability to configure an ES host
* [x] support cloud instance
* [x] abstract the `VectorMemory` logic with an adapter
  * [x] remap all the `collection` logic into ES-specific logic
  * [x] provide lean adapters for ES results
  * [ ] provide basic knn search
  * [ ] handle the switch embedder scenario

Related to issue #1039 

## Type of change

# Checklist:

### DIFF
--- a/core/cat/memory/elasticsearch/vector_memory.py
+++ b/core/cat/memory/elasticsearch/vector_memory.py
@@ -1,0 +1,62 @@
+from elasticsearch import Elasticsearch
+
+from cat.memory.constants import COLLECTION_NAMES
+from cat.memory.elasticsearch.vector_memory_collection import VectorMemoryCollectionES
+from cat.log import log
+from cat.env import get_env
+# from cat.utils import singleton
+
+
+# @singleton REFACTOR: worth it to have this (or LongTermMemory) as singleton?
+class VectorMemoryES:
+    local_vector_db = None
+
+    def __init__(
+        self,
+        embedder_name=None,
+        embedder_size=None,
+    ) -> None:
+        # connects to Qdrant and creates self.vector_db attribute
+        self.connect_to_vector_memory()
+
+        # Create vector collections
+        # - Episodic memory will contain user and eventually cat utterances
+        # - Declarative memory will contain uploaded documents' content
+        # - Procedural memory will contain tools and knowledge on how to do things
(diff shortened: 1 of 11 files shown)


core/cat/memory/elasticsearch/vector_memory.py
```diff
@@ -0,0 +1,67 @@
...
+            if es_cloud_id and len(es_cloud_id) > 0:
+                log.info(f"Connecting to Elastic Cloud: {es_cloud_id}")
+                self.vector_db = Elasticsearch(cloud_id=es_cloud_id, api_key=es_api_key)
+            else:
+                host = get_env("CCAT_ES_HOST")
+                port = get_env("CCAT_ES_PORT")
+                log.info(f"Connecting to Elastic Host: {host}:{port}")
+                self.vector_db = Elasticsearch([f"{host}:{port}"], api_key=es_api_key)
+        except AuthenticationException as err:
+                print(f"Authentication error {err=}, {type(err)=}")
```
dej611, line 54:
```suggestion
                log.error(f"Authentication error {err=}, {type(err)=}")
```

core/cat/memory/elasticsearch/vector_memory.py
```diff
@@ -0,0 +1,67 @@
...
+        except AuthorizationException as err:
+                print(f"Authorization error {err=}, {type(err)=}")
```
dej611, line 56:
```suggestion
                log.error(f"Authorization error {err=}, {type(err)=}")
```

pieroit:
In core `v2` there will be no vector memory, nor any kind of memory.
This to allow plugins to define their own DB (vector, graph, file based, or whatever) and act on memory as they wish.

Qdrant vector memory will become just a plugin, hopefully also this code can be.

Thx for your effort and sorry for being so late, had to make hard decisions

Issue #1053, opened 2025-04-01
[Feature]: refactor VectorMemoryCollection to new Qdrant APIs


nickprock:
**Is your feature request related to a problem? Please describe.**
The Qdrant version update brought several warnings for APIs that will be deprecated. The VectorMemoryCollection class needs to be updated.

**Additional context**
@pieroit if you are not already working on it I can work on it.

pieroit:
We need a major refactoring there
Is the update going to be retrocompatible?

nickprock:
> We need a major refactoring there Is the update going to be retrocompatible?

I think so, new APIs have been introduced, the only problem (which I have already run into) is some output but it should not be distruptive. If there are other changes that need to be made let's discuss them so we can put them in the queue.

nickprock:
It is complete for me, in case we can reopen it in the future 🐱

PR #1058, opened 2025-04-07
Dev/qdrant refactoring

nickprock:
# Description

Qdrant has changed the name and operation of some APIs in recent versions, in particular:
* `recreate_collection `--> `create_collection`
* `search` --> `query_points`

In the second case, the parameter names and output format have also changed. They will be deprecated in the future so it is better to update the code now without coming up short.

I have run the tests and the success rate is 100%. I also did some testing on the admin and there should be no problem, 

**I can't guarantee backward compatibility with ALL plugins.**

Related to issue #1053 

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

### DIFF
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -11,7 +11,7 @@ RUN apt-get -y update && apt-get install -y curl build-essential fastjar libmagi
 
 ### ADMIN (static build) ###
 WORKDIR /admin
-RUN curl -sL https://github.com/cheshire-cat-ai/admin-vue/releases/download/Admin/develop.zip | jar -xv
+RUN curl -sL https://github.com/cheshire-cat-ai/admin-vue/releases/download/Admin/release.zip | jar -xv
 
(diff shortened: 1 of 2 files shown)


core/Dockerfile
```diff
@@ -11,7 +11,7 @@ RUN apt-get -y update && apt-get install -y curl build-essential fastjar libmagi
 
 ### ADMIN (static build) ###
 WORKDIR /admin
-RUN curl -sL https://github.com/cheshire-cat-ai/admin-vue/releases/download/Admin/develop.zip | jar -xv
+RUN curl -sL https://github.com/cheshire-cat-ai/admin-vue/releases/download/Admin/release.zip | jar -xv
```
pieroit, line 14: did you branch off `main` instead of `develop`?
They are not in sync, all PRs should start from `develop`

nickprock: No problem, I kill the PR an submit another.

PR #1097, opened 2025-07-29
Feature/cat image only once

### LINKED ISSUE #1063: Welcome screen in terminal should appear only at first launch
This welcome appears at bootstrap and at every reload:

Image

It is awesome to see at launch, but should not appear after every reload, to ease development.

It would suffice to store a boolean flag somewhere, but not sure where.
Comment if you have ideas :)

andreaspagnolo:
# Fix: Show welcome ASCII cat only at launch

## Problem
The welcome ASCII cat art was displayed every time the application reloaded, which made development less pleasant.

## Solution
Implemented a session-based flag system that tracks whether the welcome message has been shown:

- Added `.welcome_shown` flag file to track display status
- Welcome ASCII art now only appears on first startup, not on reloads
- REST API and ADMIN URLs continue to show on every startup for convenience

### Modified `core/cat/log.py`:
- Added `_get_welcome_flag_path()` method to define flag file location  
- Added `_has_welcome_been_shown()` method to check if welcome was already displayed
- Added `_mark_welcome_as_shown()` method to mark welcome as displayed
- Updated `welcome()` method to conditionally show ASCII art based on flag
- Preserved display of REST API and ADMIN URLs on every startup

### Key Features:
- **Non-intrusive**: Flag file is created in project root and can be safely ignored
- **Developer-friendly**: Removes ASCII art clutter during development reloads
- **Preserves functionality**: Important URLs still shown on every startup
- **Clean implementation**: Uses simple file-based flag system

## Testing
- ✅ First startup shows full welcome message with ASCII cat
- ✅ Subsequent reloads show only REST API and ADMIN URLs  
- ✅ Flag file properly created and detected
- ✅ No breaking changes to existing functionality

## Files Changed
- `core/cat/log.py` - Added flag system and updated welcome method

Closes #1063

### DIFF
--- a/core/cat/log.py
+++ b/core/cat/log.py
@@ -4,6 +4,8 @@ import logging
 import sys
 import json
 import traceback
+import os
+from pathlib import Path
 from pprint import pformat
 from loguru import logger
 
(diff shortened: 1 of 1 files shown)


pieroit:
Thanks a lot @andreaspagnolo :)
Hacking on the cat is now even more pleasant!

PR #1098, opened 2025-08-01
Update embedder.py

sblendorio:
The model name was missing

# Description

Model name was missing in the configuration of openai-compatible API of embedding models


pieroit:
thanks and welcome @sblendorio

PR #1106, opened 2025-10-09
Decouple MCP types from Cat chat types

sambarza:
# Description

All Cat types related to the chat are now self-defined within Cat, and all references to MCP types are centralized in `__types_adapter.py`.
This way, if we ever want to detach from the MCP types, we only need to modify the `__types_adapter.py` class.

# Checklist:

- [X] I submitted my PR to branch `develop`
- [ ] I have run or added tests to cover my contribution

### DIFF
--- a/src/cat/types/__types_adapter.py
+++ b/src/cat/types/__types_adapter.py
@@ -1,0 +1,14 @@
+from mcp.types import Resource
+from mcp.types import TextContent
+from mcp.types import ImageContent
+from mcp.types import AudioContent
+from mcp.types import ResourceLink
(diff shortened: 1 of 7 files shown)


pieroit:
thanks @sambarza, will follow the same pattern for typing and subclassing external classes

pieroit:
Centralized the types for external import [here](https://github.com/cheshire-cat-ai/core/blob/v2/src/cat/types/__init__.py) so we can internally do as we please

From plugins and other core components, you just do:
```python
from cat.types import Context, Resource, Message, ImageBlock # etc
```

Issue #1109, opened 2025-11-28
Tools as class methods


pieroit:
### Intro

Help needed on designing and implementing a truly flexible tool system for Cat v2 agents.

### Background

In `v2` tools will come from three different places:

1 - MCP servers
2 - global tools (defined in plugins) with the `@tool` decorator
3 - agent specific tools (defined in plugins, directly in the agent class)

1 and 2 already work, this issue is on 3.

### Agent specific tools

Tools available to an agent come from the `agent.list_tools()` method. The method gives out a list of [`CatTool`](https://github.com/cheshire-cat-ai/core/blob/v2/src/cat/mad_hatter/decorators/tool.py) objects to be passed to `agent.llm(...)` alongside messages and other params. Example of the typical agentic loop is found in the [default agent](https://github.com/cheshire-cat-ai/core/blob/9974f71ca77241058fdedb832495a2081f7a85ec/src/cat/agents/default.py#L8):

```python
        while True:
            llm_mex: Message = await self.llm(
                # delegate prompt construction to plugins
                await self.get_system_prompt(),
                # pass conversation messages
                messages=self.chat_request.messages + self.chat_response.messages,
                # pass tools (both internal and MCP)
                tools=await self.list_tools(),
                # whether to stream or not
                stream=self.chat_request.stream,
            )

            # do stuff with the message
```

I'll probably wrap this loop in a method like `agent.inner_loop()` so it can be reused or overridden at your pleasure. The issue here regards `list_tools`.

### How `list_tools` work

The method is included in `BaseAgent`:

```python
    async def list_tools(self) -> List[CatTool]:
        """Get both plugins' tools and MCP tools in CatTool format."""

        mcp_tools = await self.mcp.list_tools()
        mcp_tools = [
            CatTool.from_fastmcp(t, self.mcp.call_tool)
            for t in mcp_tools
        ]

        tools = await self.execute_hook(
            "agent_allowed_tools",
            mcp_tools + self.mad_hatter.tools
        )

        return tools
```

What is missing here are tools defined in the agent itself, in the _most simple way possible_.
Something like (in a plugin):

```python
from cat import tool, BaseAgent

class MyAgent(BaseAgent):

    async def execute(self):
        # ...
    
    @tool
    async def get_weather(self, city: str):
        # possibility here to use `self.xxx`
        return x
```

[...]

PR #1111, opened 2025-12-15
[V2] Class based agent tools 

lucagobbi:
# Description

This PR adds the ability to define tools directly as methods on Agent classes using the `@agent_tool` decorator. Highly inspired by [SuperCatForm's class based tools](https://github.com/lucagobbi/super-cat-form) using Python `__get__` descriptor to bind the tool to the owning instance.

Mentioned #1109 

### Examples

```python
class MyAgent(Agent):
    
    @agent_tool
    async def search(self, query: str) -> str:
        """Search the web for information."""
        # Full access to self.cat, agents state and attributes, etc.
        return await self.some_api.search(query)
    
    @agent_tool(return_direct=True)
    def get_user_name(self) -> str:
        """Get the current user's name."""
        return self.user.name
```

@pieroit couldn't test it deeply 'cause I'm not yet practical with v2, but this could be a starting point.

### DIFF
--- a/src/cat/agents/base.py
+++ b/src/cat/agents/base.py
@@ -255,17 +255,25 @@ class Agent(Service, LLMMixin, EventStreamMixin):
 
         for name, func in self.get_agent_tools().items():
             # Create a bound method wrapper that includes self
-            def create_bound_wrapper(bound_func):
+            sig = inspect.signature(func)
+            valid_params = set(sig.parameters.keys()) - {'self'}
+
+            def create_bound_wrapper(bound_func, valid_params):
+
                 async def wrapper(**kwargs):
+                    filtered_kwargs = {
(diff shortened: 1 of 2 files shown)


lucagobbi:
@pieroit I reasoned about our discussion last night and I have come to the conclusion that the two separated decorators with the same name are a pain in d ass. So I tried to simplify it much further keeping one single tool decorator and adding the `bind_to` method to the Tool class allowing you to bind the class based tool to the specific instance (could also be used also in other circumstances rather than Agents)

In this way agents declare its own tools with the same decorator binding em to its instance and that's it. Hope you like it.

Btw, as I've said, I couldnt test it throughly because my v2 is really messed up.

pieroit:
Thanks @lucagobbi

PR #1133, opened 2026-06-23
fix: load bundled plugins on Windows

davidegavio:
On Windows, `os.path.relpath` returns paths with backslash separators (e.g. `chats\db.py`). The old conversion (`.replace("/", ".")`) only handled forward slashes, producing module names like `chats\db` that `importlib.import_module` cannot resolve.

Replace brittle `.replace(".py", "")` with `os.path.splitext` and use `os.sep` so the platform-native separator is always converted to the dotted notation required by Python's import system.


pieroit:
awesome thx!

PR #1143, opened 2026-07-29
feat: support AG-UI input and client tools

### LINKED ISSUE #1142: Support AG-UI RunAgentInput as optional input format (bidirectional AG-UI)
## Current behavior  
  
The framework currently supports AG-UI only in one direction: it emits AG-UI events  (SSE) from server to client, but does not accept AG-UI events as input.  Client input always goes through a proprietary `Task` object via HTTP POST.  
  
This means that native AG-UI clients (which send a `RunAgentInput` payload) cannot  talk directly to the framework — an adapter layer is required to translate  `RunAgentInput` → `Task` before the request reaches the agent endpoint.  
  
## Proposed change  
  
Add optional support for accepting `RunAgentInput` as an alternative input format  on the agent message endpoint (or a dedicated endpoint), so that native AG-UI clients  can interact with the framework without needing an external adapter.  
  
The existing `Task`-based input should remain the default and be fully supported —  this would be purely additive/opt-in.  
  
## Motivation  
  
AG-UI is already a first-class citizen on the output side. Supporting it on the input  side as well would make the integration story for native AG-UI clients much simpler,  removing the need for a proxy/adapter layer just to bridge the input format gap.  
  
The mapping between `RunAgentInput` and `Task` is relatively straightforward:  
- `RunAgentInput.messages` → `Task.messages` (both follow the OpenAI message format)  
- `RunAgentInput.thread_id` / `run_id` → currently unused in `Task` (there is already  
  a `TODOV2` comment in `AGUIStream.__init__` noting that `thread_id` should come  from the request/session)  
- `RunAgentInput.tools`, `context`, etc. → `Task.resources` / `Task.args`  
  
## References  
  
- AG-UI protocol: https://github.com/ag-ui-protocol/ag-ui  
- Current streaming implementation: `src/cat/protocols/agui/streaming.py`  
- Agent endpoint: `src/cat/routes/agents.py`

serrapa:
## Summary

- accept standard AG-UI `RunAgentInput` on the agent message endpoint
- stream compliant AG-UI events through SSE
- preserve non-streaming `Task` support
- expose client-provided tools to the model without executing them server-side
- continue frontend tool calls through a subsequent `ToolMessage`
- validate tool names, schemas, limits and server-tool collisions

## Testing

- Ruff passed
- 61 focused tests passed
- verified the endpoint through a live Uvicorn instance
- verified a complete frontend-tool round trip over HTTP/SSE

The full suite on Windows reports unrelated existing failures around ZIP MIME detection and POSIX path assertions: 211 passed, 11 failed, 28 errors.

## Security

Client tools are runtime-only and have no server callable. Normal `Task` payloads cannot inject them, server tool names cannot be shadowed, and request definitions have resource limits.

Closes #1142


pieroit:
thx will need time to check this one. Appreciated

